PR #7, opened 2025-03-04
add cranked txns to sankey

jherrera-jump:


atsai-jumptrading:
I think you might need to change line 128 of SlotSankey.tsx to include `waterfall.in.pack_cranked` check the numbers in the sankey just to make sure 
```
 const bankCount =
    waterfall.in.pack_retained + packCount - getPackOut(waterfall.out);
```

atsai-jumptrading (approved):

PR #8, opened 2025-03-05
add indicator for first processed slot

jherrera-jump:
Adds green flag icon to epoch bar to indicate first slot not included in the incremental snapshot (i.e. the first "processed" slot)

### DIFF
--- a/src/api/entities.ts
+++ b/src/api/entities.ts
@@ -250,11 +250,6 @@ export const skipRateSchema = z.object({
   // slots_skipped: z.number(),
 });
 
-export const firstProcessedSlotSchema = z.object({
-  epoch: z.number(),
-  slot: z.number().nullable(),
-});
-
 export const summarySchema = z.discriminatedUnion("key", [
   summaryTopicSchema.extend({
     key: z.literal("ping"),
@@ -344,10 +339,6 @@ export const summarySchema = z.discriminatedUnion("key", [
   summaryTopicSchema.extend({
     key: z.literal("skip_rate"),
     value: skipRateSchema,
-  }),
-  summaryTopicSchema.extend({
-    key: z.literal("first_processed_slot"),
-    value: firstProcessedSlotSchema,
   }),
 ]);
 
--- a/src/features/EpochBar/EpochSlider.tsx
+++ b/src/features/EpochBar/EpochSlider.tsx
@@ -13,12 +13,15 @@ import { Box } from "@radix-ui/themes";
 import { atom, useAtom, useAtomValue, useSetAtom } from "jotai";
 import { skippedSlotsAtom } from "../../api/atoms";
 import warning from "../../assets/warning_16dp_FF5353_FILL1_wght400_GRAD0_opsz20.svg";
+import green_flag from "../../assets/flag.svg";
 import {
   currentLeaderSlotAtom,
   slotOverrideAtom,
   leaderSlotsAtom,
   epochAtom,
+  firstProcessedSlotAtom,
 } from "../../atoms";
+import { startupProgressAtom } from "../../api/atoms";
 import { useInterval, useMeasure } from "react-use";
 import { Epoch } from "../../api/types";
 
@@ -42,6 +45,7 @@ function slotToEpochPct({
   )
     return 0;
 
+  slot = Math.min(Math.max(slot, epochStartSlot), epochEndSlot);
   const totalEpochSlots = epochEndSlot - epochStartSlot;
   const epochSlotProgress = slot - epochStartSlot;
   return epochSlotProgress / totalEpochSlots;
@@ -122,6 +126,7 @@ function EpochSlider({ canChange }: EpochSliderProps) {
   const [slotOverride, setSlotOverride] = useAtom(slotOverrideAtom);
   const leaderSlots = useAtomValue(leaderSlotsAtom);
   const skippedSlots = useAtomValue(skippedSlotsAtom);
+  const firstProcessedSlot = useAtomValue(firstProcessedSlotAtom);
   const [value, setValue] = useState(() => {
     return [
       pctToValue(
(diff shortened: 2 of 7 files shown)


src/features/LeaderSchedule/Slots/SlotCardGrid.tsx
```diff
@@ -176,6 +177,8 @@ interface SlotCardRowProps {
 }
 
 function SlotCardRow({ slot, active }: SlotCardRowProps) {
+  const startupProgress = useAtomValue(startupProgressAtom);
+  const firstProcessedSlot = ((startupProgress?.downloading_incremental_snapshot_slot ?? -1) + 1) || undefined;
```
atsai-jumptrading, line 181: You could just add an atom for this since you're having to repeat. This would also make the consumers not re-render when other parts of `startupProgress` changes. Not really necessary though what you have here is fine too 

```
const firstProcessedSlotAtom = atom((get) => {
const startupProgress = useAtomValue(startupProgressAtom);
if(!startupProgress?.downloading_incremental_snapshot_slot ) return;

return startupProgress.downloading_incremental_snapshot_slot +1; 
})
```

and consumers would just have 
```
const firstProcessedSlot = useAtomValue(firstProcessedSlotAtom)
```

src/features/EpochBar/EpochSlider.tsx
```diff
@@ -279,7 +312,7 @@ function LeaderSlot({ slot, pct, width }: LeaderSlotProps) {
       className={styles.leaderSlot}
       style={{
         left: `${pct * 100}%`,
-        background: isFutureSlot ? "rgba(255, 255, 255, 0.20)" : undefined,
+        background: isFutureSlot || isSnapshotSlot ? "rgba(255, 255, 255, 0.20)" : undefined,
```
atsai-jumptrading, line 315: This is just gonna color everything after the snapshot slot grey isn't it? Even if it's not in the future? `return slot >= (firstProcessedSlot ?? 0);`

src/features/EpochBar/EpochSlider.tsx
```diff
@@ -147,6 +150,18 @@ function EpochSlider({ canChange }: EpochSliderProps) {
     slotToEpochPct
   );
 
+  const startupProgress = useAtomValue(startupProgressAtom);
+  const firstProcessedSlot = ((startupProgress?.downloading_incremental_snapshot_slot ?? -1) + 1) || undefined;
+  const [epochSnapshotPct, updateEpochSnapshotPct] = useReducer(
```
atsai-jumptrading, line 155: You shouldn't need this whole reducer because we shouldn't need to manually update it. `epochProgressPct` is always changing as the epoch progresses, so it needs to update at an interval to keep the bar live while not constantly re-rendering. `firstProcessedSlot` is a static slot and will only update when the validator gets restarted. To get the pct of it we should just follow the skipped slots example that doesn't use the reducer and a `useMemo` instead. `skippedSlotPcts`

src/features/EpochBar/EpochSlider.tsx
```diff
@@ -155,6 +170,11 @@ function EpochSlider({ canChange }: EpochSliderProps) {
           epochStartSlot: epoch?.start_slot,
           epochEndSlot: epoch?.end_slot,
         });
+        updateEpochSnapshotPct({
```
atsai-jumptrading, line 173: This is also unneeded then since we don't need to update at an interval

src/features/EpochBar/EpochSlider.tsx
```diff
@@ -242,6 +262,9 @@ function EpochSlider({ canChange }: EpochSliderProps) {
         {skippedSlotPcts?.map(({ slot, pct }) => (
           <SkippedSlot key={slot} slot={slot} pct={pct} />
         ))}
+        {firstProcessedSlot
+          ? <FirstProcessedSlot key={firstProcessedSlot} slot={firstProcessedSlot} pct={epochSnapshotPct}/>
```
atsai-jumptrading, line 266: No need a key for this as it's not in a map. Can also just do `firstProcessedPct && <FirstProcessedSlot...`

src/api/entities.ts
```diff
@@ -345,10 +340,6 @@ export const summarySchema = z.discriminatedUnion("key", [
     key: z.literal("skip_rate"),
     value: skipRateSchema,
   }),
-  summaryTopicSchema.extend({
```
jherrera-jump, line 348: I can re-add these if they were needed, but they seem to not be used?

atsai-jumptrading: Oh yea they were how we were sending the msg before we removed it. Not needed

src/features/EpochBar/EpochSlider.tsx
atsai-jumptrading: I don't think you need to change this component at all though? This is the `LeaderSlot` ticks and the first slot isn't necessarily a leader slot. The tick and coloring and all should be there with the `FirstProcessedSlot` component too so shouldn't need to modify this or have an `isSnapshotSlotAtom`

src/features/EpochBar/EpochSlider.tsx
```diff
@@ -242,6 +256,7 @@ function EpochSlider({ canChange }: EpochSliderProps) {
         {skippedSlotPcts?.map(({ slot, pct }) => (
           <SkippedSlot key={slot} slot={slot} pct={pct} />
         ))}
+        {firstProcessedSlotPct && firstProcessedSlot && <FirstProcessedSlot slot={firstProcessedSlot} pct={firstProcessedSlotPct}/>}
```
atsai-jumptrading, line 259: Also just in case are for some reason 0, `firstProcessedSlotPct` and `firstProcessedSlot`, will evaluate to false but show a `0` on the page. So need to do `!!firstProcessedSlotPct` or `Boolean(firstProcessedSlotPct )`. 

The other option also that I forgot to mention is just make these props nullable or undefined as well whichever one is needed, then just put a guard at the top `if(!firstProcessedSlot || !firstProcessedSlotPct) return` makes this part slightly cleaner. Up to you though that's preference

jherrera-jump:
> I don't think you need to change this component at all though? This is the LeaderSlot ticks and the first slot isn't necessarily a leader slot. The tick and coloring and all should be there with the FirstProcessedSlot component too so shouldn't need to modify this or have an isSnapshotSlotAtom

yea I'll take this out. I think currently snapshot slots are shown as not skipped always, since the frontend doesn't know which ones were actually skipped -- but after thinking about it this isn't a great way to let the user know that.

atsai-jumptrading (approved):

PR #16, opened 2025-03-24
change timestamps + lamports fields to string

jherrera-jump:


src/features/Gossip/Grid.tsx
```diff
@@ -48,14 +49,14 @@ const colDefs: ColDef<PeerRow>[] = [
...
-    cellStyle: (params) => {
+    cellStyle: (params: ValueFormatterParams<PeerRow, string>) => {
       return { color: params.data?.delinquent ? "#6D6F71" : "#B2BCC9" };
     },
     filter: true,
   },
   { field: "name", initialWidth: 300, filter: true },
   { field: "description", initialWidth: 500, filter: true },
-  { field: "stake", initialSort: "desc", initialWidth: 180, filter: true },
+  { field: "stake", initialSort: "desc", initialWidth: 180, filter: true, valueFormatter: (params: ValueFormatterParams<PeerRow, bigint>) => params.value?.toString() ?? "" },
```
atsai-jumptrading: Is this needed? Do bigints not display properly on the grid?

src/utils.ts
```diff
@@ -76,14 +75,16 @@ export const fixValue = (val: number) =>
   val >= 18446744073709552000 ? 0 : val;
 
 export function getStake(peer: Peer) {
-  return sum(peer.vote.map(({ activated_stake }) => activated_stake));
+  return peer.vote
+    .map(({ activated_stake }) => activated_stake)
+    .reduce((acc, num) => acc + num, BigInt(0));
```
atsai-jumptrading, line 80: Why change this from the sum? Does it not take bigint? Also doesn't really matter here, but the lodash sum does measure faster than the reduce on large arrays. If it does need to be changed, then no need to also map still. Just have a reduce ` .reduce((total, {activated_stake}) => total + activated_stake, BigInt(0));`

src/utils.ts
```diff
@@ -76,14 +75,16 @@ export const fixValue = (val: number) =>
...
 }
 
-export function getFmtStake(stake?: number) {
+export function getFmtStake(stake?: bigint) {
   if (stake === undefined) return;
 
   let value = "";
-  const solAmount = stake / lamportsPerSol;
+  const solAmount = Number(stake) / lamportsPerSol;
```
atsai-jumptrading, line 87: I guess for stuff like this we're ok just losing slight precision right. We should eventually add a divider helper function that will divide based on decimals we want precisely

src/features/Overview/ValidatorsCard/formatAmt.ts
```diff
@@ -2,13 +2,13 @@ import { lamportsPerSol } from "../../../consts";
...
 
 export function formatNumberLamports(
-  value: number,
+  value: bigint,
   decimalCount: number = 3,
-  formatOptions?: FormatNumberOptions
+  formatOptions?: FormatNumberOptions,
 ) {
-  if (value === 0) return "0";
+  if (value === BigInt(0)) return "0";
```
atsai-jumptrading, line 9:
```suggestion
  if (!value) return "0";
```

src/features/Overview/ValidatorsCard/formatAmt.ts
```diff
@@ -2,13 +2,13 @@ import { lamportsPerSol } from "../../../consts";
 import { formatNumber, FormatNumberOptions } from "../../../numUtils";
 
 export function formatNumberLamports(
-  value: number,
```
atsai-jumptrading, line 5: Should probably eventually also make one of these for numbers vs bigints too, so we're not doing this bigint number casting everywhere

atsai-jumptrading (approved):

src/features/Gossip/Grid.tsx
jherrera-jump: I was getting console warning since I guess AG Grid doesn't have a built-in formatter for bigint--don't remember if they actually displayed, will check

src/utils.ts
jherrera-jump: sum returns number, so changed to get rid of the implicit cast

atsai-jumptrading: Ah makes sense

src/atoms.ts
```diff
@@ -419,9 +419,9 @@ export const myStakePctAtom = atom((get) => {
   const peerStats = get(peerStatsAtom);
   const stake = get(myStakeAmountAtom);
 
-  if (stake === undefined || !peerStats) return;
+  if (stake === undefined || !peerStats || !(peerStats.activeStake || peerStats.delinquentStake)) return;
```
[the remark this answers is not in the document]
jherrera-jump: Do you mean `(!peerStats.activeStake && !peerStats.delinquentStake)`? This was to prevent divide by 0.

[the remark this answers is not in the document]
atsai-jumptrading: oh nvm i completely misread 😅 this is fine. I personally would just add another line and do `if(!peerStats.activeStake && !peerStats.delinquentStake) return;` since I think it's clearer to have multiple guard lines vs nesting parens. Either way works here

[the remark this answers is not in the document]
atsai-jumptrading: Yup misread initially that's what i meant

atsai-jumptrading (approved):

atsai-jumptrading (dismissed):

PR #19, opened 2025-04-04
fix: support websocket url configuration

amliu-jumptrading:
# fix: support websocket url configuration

I have added support for configuring the WebSocket URL to simplify swapping it during development. After considering two approaches — using an `.env` file or a configuration file — I chose to use the .env file for the following reasons:

It suits our needs for storing simple key-value pairs.
The WebSocket URLs are runtime-specific and should not be committed to source control.
While configuration files are better suited for structured data (e.g., JSON) or scenarios where settings need to be committed, the `.env` file approach was a better fit for our use case.

Additionally, I’ve specified in the `README.md` for developers to create a .env.local file. This is useful because `.env.local` is ignored by `npm build` in most frameworks that use environment files (including `Vite`). It allows developers to define local overrides without affecting production builds or committing sensitive information.


src/api/ws/ConnectionProvider.tsx
```diff
@@ -12,10 +12,14 @@ import {
   messageEventType,
 } from "./ConnectionContext";
 
+const VITE_WEBSOCKET_URL = import.meta.env.VITE_WEBSOCKET_URL as string;
```
atsai-jumptrading, line 15: do we need to trim or are spaces when ignored from the file?

atsai-jumptrading (approved):

PR #37, opened 2025-06-24
gui: send txn counts, boot timestamp

jherrera-jump:


src/features/LeaderSchedule/Slots/SlotCardGrid.tsx
```diff
@@ -324,7 +325,7 @@ function SlotCardRow({ slot, active }: SlotCardRowProps) {
         className={`${styles.rowText} ${active ? styles.active : ""}`}
         align="right"
       >
-        {getText(values?.nonVoteTxns)}
+        {getText(values?.userTxns)}
```
atsai-jumptrading, line 328: do we want to do any renaming on the labels to user txn?

src/features/Overview/SlotPerformance/SlotSankey/index.tsx
```diff
@@ -359,8 +359,16 @@ function SlotSankey({ slot }: { slot?: number }) {
       waterfall,
       displayType,
       query.response?.publish.duration_nanos,
-      query.response?.publish.transactions,
-      query.response?.publish.failed_transactions,
+      query.response?.publish.success_user_transaction_cnt &&
+        query.response?.publish.success_vote_transaction_cnt
+        ? query.response?.publish.success_user_transaction_cnt +
+            query.response?.publish.success_vote_transaction_cnt
```
atsai-jumptrading, line 365: in `getLinks` there is a `success = totalTransactions - failedTransactions`. Here `totalTransactions` is now only `success_vote + success_user` not including failed. Is that correct?

src/features/Overview/SlotPerformance/SankeyControls.tsx
```diff
@@ -119,6 +130,11 @@ function SlotStats() {
 
   if (!selectedSlot) return;
 
+  // right-align txn counts by inserting invisible numbers as padding
```
atsai-jumptrading, line 133: It looks like you're trying to left align by putting invisible text afterwards? Why do we want a different alignment than how it is currently? Think we can probably get it aligned without invisible text too

src/api/atoms.ts
```diff
@@ -34,8 +33,8 @@ export const commitHashAtom = atom<CommitHash | undefined>(undefined);
 
 export const identityKeyAtom = atom<IdentityKey | undefined>(undefined);
 
-export const uptimeAtom = atom<
-  { uptimeNanos: UptimeNanos; ts: DateTime } | undefined
+export const startupTimeAtom = atom<
+  { startupTimeNanos: StartupTimeNanos } | undefined
```
atsai-jumptrading, line 37: So it now sends a constant start time vs a moving uptime?

src/features/Overview/SlotPerformance/SlotSankey/index.tsx
jherrera-jump: yea your're right

src/features/Overview/SlotPerformance/SankeyControls.tsx
```diff
@@ -153,9 +164,20 @@ function SlotStats() {
         <RowSeparator my="0" />
       </div>
       <Text>Vote Transactions</Text>
-      <Text align="right">{values?.voteTxns?.toLocaleString() ?? "-"}</Text>
+      <Text align="right">
+        {(values?.failedVoteTxns != null && values?.successfulVoteTxns != null
```
atsai-jumptrading, line 168: I think it's just `values` that can be undefined? Not the actual values `failedVoteTxns` and `successfulVoteTxns`. So just check for `values ? ... : undefined`. Since how this looks is that if only one of the 2 are null then it'll just not show a value vs showing the remaining. Same below for nonvote

src/features/Overview/SlotPerformance/SlotSankey/index.tsx
```diff
@@ -359,8 +359,20 @@ function SlotSankey({ slot }: { slot?: number }) {
       waterfall,
       displayType,
       query.response?.publish.duration_nanos,
-      query.response?.publish.transactions,
-      query.response?.publish.failed_transactions,
+      query.response?.publish?.success_nonvote_transaction_cnt != null &&
+        query.response?.publish?.success_vote_transaction_cnt != null &&
+        query.response?.publish?.failed_vote_transaction_cnt &&
+        query.response?.publish?.failed_nonvote_transaction_cnt
```
atsai-jumptrading, line 365: Do need to check for `!= null` vs just truthy since they can be 0, same below. Just var out the failed above it'll read clearer `const failedTransactions = ...`

PR #40, opened 2025-07-15
[Fix] Bank chart tooltip txn sig copy button to fallback when navigator.clipboard is not available 

atsai-jumptrading:


src/utils.ts
```diff
@@ -111,3 +111,32 @@ export const hasModKey = ({
...
+  copyEl.style.position = "absolute";
+  copyEl.style.left = "-999999px";
+
+  document.body.appendChild(copyEl);
+  copyEl.select();
+
+  try {
+    const successful = document.execCommand("copy");
+    if (!successful) {
+      console.error("Failed to copy text");
```
jherrera-jump, line 135: can you also log copyValue

jherrera-jump (approved):

asuzuki-jumptrading (approved):

PR #54, opened 2025-08-15
feat: slot navigation filter toggles

asuzuki-jumptrading:
- add My Slots filter for slot navigation
- default filter state is All Slots
- navigating to the Leader Schedule will reset the selected nav filter state

### DIFF
--- a/src/consts.ts
+++ b/src/consts.ts
@@ -1,6 +1,5 @@
 export const slotsPerLeader = 4;
-export const slotsListPinnedSlotOffset = 1;
-export const slotsListFutureSlotsOffset = 5;
+export const slotsListPinnedSlotOffset = 5;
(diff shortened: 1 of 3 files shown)


src/atoms.ts
```diff
@@ -178,6 +186,13 @@ export const deleteSlotStatusBoundsAtom = atom(null, (get, set) => {
           continue;
         }
 
+        if (
```
amliu-jumptrading, line 189: In my slots mode, are we keeping the statuses for all leader slots? Would this have a memory impact on validators with a lot of leader slots during the epoch? This seems like it could be difficult, so feel free to push back, but could we have an atom that tracks how many slots are visible in the slots override and keep the statuses for the leader slots within some bounded based on that visible number of slots value?

atsai-jumptrading (approved):

PR #57, opened 2025-08-19
fix: move epoch check to selected slot getter

asuzuki-jumptrading:
- selected slot check needs to pick up changes to the epoch
- don't use `initialTopMostItemIndex` because it clashes with the scroll to `slotOverride`
- don't scroll within an atom (the slotOverride setter): this causes race conditions where the list scrolls first during atom setting, but the list item styling don't pick up changes until the next render

### DIFF
--- a/src/features/Navigation/SlotsList.tsx
+++ b/src/features/Navigation/SlotsList.tsx
@@ -25,6 +25,9 @@ import { useCurrentRoute } from "../../hooks/useCurrentRoute";
 
 const computeItemKey = (slot: number) => slot;
 
+// Add one future slot to prevent current leader transition from flickering
+const increaseViewportBy = { top: 24, bottom: 0 };
+
 interface SlotsListProps {
   width: number;
   height: number;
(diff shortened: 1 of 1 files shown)


src/features/Navigation/SlotsList.tsx
```diff
@@ -141,36 +118,22 @@ function InnerSlotsList({
       container.removeEventListener("wheel", handleScroll);
       container.removeEventListener("touchmove", handleScroll);
     };
-  }, [getSlotAtIndex, setSlotOverride, slotsCount, visibleStartIndexRef]);
-
-  // Auto scroll enabled (live scrolling)
```
asuzuki-jumptrading, line 146: Moved to `RTAutoScroll` component

src/features/Navigation/SlotsList.tsx
```diff
@@ -55,51 +66,15 @@ function InnerSlotsList({
...
-  const selectedSlot = useAtomValue(selectedSlotAtom);
-  const currentLeaderSlot = useAtomValue(currentLeaderSlotAtom);
-
   const setSlotOverride = useSetAtom(slotOverrideAtom);
-  const autoScroll = useAtomValue(autoScrollAtom);
-  const setSlotScrollListFn = useSetAtom(setSlotScrollListFnAtom);
-
   const slotsCount = slotGroupsDescending.length;
 
-  useEffect(() => {
```
asuzuki-jumptrading, line 67: Moved to `SlotOverrideScroll` component

PR #61, opened 2025-08-27
fix: detail page navigation

amliu-jumptrading:
# fix: detail page navigation


src/app.css
```diff
@@ -38,7 +37,7 @@
 
 .sticky {
   position: sticky;
-  z-index: 3;
+  z-index: 1000;
```
atsai-jumptrading, line 40: Don't need to do this now but we should move all the z-index's to vars like the colors so they're organized/easier to look at a whole. Some tooltips might be hidden by setting the z-index so high here

src/components/SlotClient.tsx
```diff
@@ -5,13 +5,13 @@ import { memo } from "react";
...
   slot,
-  size = "11px",
+  size = "12px",
 }: {
   slot: number;
   size?: string;
 }) {
   const { client } = useSlotInfo(slot);
-  if (!client) return;
+  if (!client) return <div style={{ height: size, padding: "1px" }} />;
```
atsai-jumptrading, line 14: let's just remove the size prop, we don't need it to be so dynamic. Just use a flag size prop that sets a class with a set height so we can remove the inline style

src/components/StatusIcon.tsx
```diff
@@ -93,4 +95,20 @@ export const LoadingIcon = ({
       />
     </Flex>
   );
-};
+}
+
+export function SkippedIcon({ iconSize = "14px" }: { iconSize?: string }) {
```
atsai-jumptrading, line 100: Same here for this one since it's repeated a lot in the leader schedule. Or if we don't want 2 classes just use a class that uses a css var, and have the `iconSize` be a css var that's set inline

src/features/Navigation/SlotsRenderer.tsx
```diff
@@ -63,65 +70,57 @@ const MSlotsRenderer = memo(function SlotsRenderer({
...
-              [styles.nextSlotGroup]: isNextLeader,
-              [styles.previousSlotGroup]: isPreviousLeader,
-              [styles.futureSlotGroup]: isFutureLeader,
-              [styles.pastSlotGroup]: isPastLeader,
-              [styles.selectedSlotGroup]: isSelectedLeader,
-              [styles.skippedSlotGroup]: hasSkipped,
-              [styles.yourSlotGroup]: isYou,
-            })}
+            direction="column"
+            overflow="hidden"
```
atsai-jumptrading, line 97: i see you just moved this so it's fine but it feels off that this is needed

src/features/SlotDetails/index.tsx
```diff
@@ -112,24 +112,27 @@ function SlotSearch() {
...
-          </IconButton>
-        </TextField.Slot>
-      </TextField.Root>
+      <Flex direction="column" width="100%" maxWidth="510px">
+        <TextField.Root
+          placeholder={`Slot ID e.g. ${epoch?.start_slot ?? 0}`}
+          type="number"
+          step="1"
+          value={searchSlot}
+          onKeyDown={handleKeyDown}
```
atsai-jumptrading, line 125: we should just wrap the input in a form to handle the enter vs explicit onKeyDown

src/features/SlotDetails/index.tsx
```diff
@@ -258,77 +262,91 @@ function SlotHeader({ slot }: { slot: number }) {
...
+    </Flex>
+  );
+}
+
+function SlotHeader({ slot }: { slot: number }) {
+  const { peer, isLeader, name } = useSlotInfo(slot);
+
+  return (
+    <Flex gap="3" wrap="wrap" className={styles.header}>
+      <Flex gap="3" align="center">
```
atsai-jumptrading, line 331: Is the extra `Flex` necessary? Feel like we shouldn't need an outer flex just for 1 child

src/features/SlotDetails/slotDetails.module.css
```diff
@@ -34,42 +37,81 @@
...
+
+    &.selected-slot {
+      font-weight: 600;
+      background: #283551;
+      color: #e3efff;
+    }
+
+    &.skipped-slot {
+      background: #250f0f;
+      color: #9aabc3;
```
atsai-jumptrading, line 78: let's pull the colors into the colors file if possible for organization

src/features/SlotDetails/slotDetails.module.css
```diff
@@ -34,42 +37,81 @@
...
+  & .link {
+    text-decoration: none;
+    flex: 1;
   }
 
-  &.skipped-slot.selected-slot {
-    border-color: rgba(255, 71, 71, 0.65);
-    background-color: #3b0c0c;
+  & .slot-status {
+    cursor: pointer;
```
atsai-jumptrading, line 58: This is used for the links it looks like? Shouldn't the links already be pointers? Not sure why we need to explicitly add an extra pointer unless we're just trying to get the padding or something like that

src/features/SlotDetails/slotDetails.module.css
```diff
@@ -34,42 +37,81 @@
...
+    }
+
+    &.skipped-slot.selected-slot {
+      background: #3b0c0c;
+    }
+  }
+
+  &.previous-navigation,
+  &.next-navigation {
+    overflow: hidden;
```
atsai-jumptrading, line 88: is the overflow:hidden needed?

src/features/SlotDetails/index.tsx
```diff
@@ -24,26 +19,31 @@ import {
...
 import { useSlotQueryPublish } from "../../hooks/useSlotQuery";
 import { getLeaderSlots, getSlotGroupLeader } from "../../utils";
-import { StatusIcon } from "../../components/StatusIcon";
+import { SkippedIcon, StatusIcon } from "../../components/StatusIcon";
 import clsx from "clsx";
 import { failureColor } from "../../colors";
 
 import skippedIcon from "../../assets/Skipped.svg";
-import green_flag from "../../assets/flag.svg";
+import earliest from "../../assets/earliest.svg";
```
asuzuki-jumptrading, line 34: I think we may be able to reuse the `history.svg` icon

src/features/SlotDetails/index.tsx
```diff
@@ -112,24 +112,27 @@ function SlotSearch() {
...
-        onChange={(e) => setSearchSlot(e.target.value)}
-        size="3"
-      >
-        <TextField.Slot side="right">
-          <IconButton variant="ghost" color="gray" onClick={submitSearch}>
-            <MagnifyingGlassIcon height="16" width="16" />
-          </IconButton>
-        </TextField.Slot>
-      </TextField.Root>
+      <Flex direction="column" width="100%" maxWidth="510px">
```
asuzuki-jumptrading, line 119: Could we use a regular `div` for this?

src/features/SlotDetails/index.tsx
```diff
@@ -223,7 +226,7 @@ function QuickSearchCard({
       </Text>
       <Flex width="100%" justify="between">
         <Text size="1" style={{ color: "#cecece" }}>
-          {slot}
+          {slot !== undefined ? slot : "--"}
```
asuzuki-jumptrading, line 229: nit:
`slot ?? "--"`

src/features/SlotDetails/index.tsx
amliu-jumptrading: I've removed it

src/features/SlotDetails/slotDetails.module.css
amliu-jumptrading: I was able to remove it

src/assets/Skipped.svg
```diff
@@ -1,3 +1,3 @@
-<svg width="18" height="18" viewBox="0 0 18 18" fill="none" xmlns="http://www.w3.org/2000/svg">
-<path fill-rule="evenodd" clip-rule="evenodd" d="M8.01469 6.00051C5.58562 6.03619 3 7.94365 3 12C3 12.5523 2.55228 13 2 13C1.44772 13 1 12.5523 1 12C1 6.90007 4.41438 4.05318 7.98531 4.00073C9.74494 3.97489 11.5153 4.63461 12.8432 6.00502C13.8812 7.07635 14.6101 8.5405 14.8821 10.3602L15.8896 8.9389C16.209 8.48833 16.8331 8.38198 17.2837 8.70137C17.7343 9.02075 17.8406 9.64492 17.5213 10.0955L15.0678 13.5568C14.7484 14.0073 14.1242 14.1137 13.6737 13.7943L10.2124 11.3408C9.76181 11.0214 9.65546 10.3973 9.97484 9.9467C10.2942 9.49613 10.9184 9.38978 11.369 9.70916L12.9263 10.8131C12.7275 9.27771 12.1449 8.15845 11.4068 7.39676C10.4847 6.44507 9.25506 5.9823 8.01469 6.00051ZM6.5 13C7.32843 13 8 12.3284 8 11.5C8 10.6716 7.32843 10 6.5 10C5.67157 10 5 10.6716 5 11.5C5 12.3284 5.67157 13 6.5 13Z" fill="#D86363"/>
+<svg width="18" height="18" viewBox="0 0 18 18" fill="#D86363" xmlns="http://www.w3.org/2000/svg">
```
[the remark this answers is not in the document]
amliu-jumptrading: If the fills are in the individual paths, you cannot override it with the `fill` field you pass into the react svg component. You can only override the fill field of the top level svg

atsai-jumptrading (approved):

asuzuki-jumptrading (approved):
Nice, thank you!

PR #99, opened 2025-10-07
feat: add quick select list options

amliu-jumptrading:


src/features/SlotDetails/SlotSearch.tsx
```diff
@@ -0,0 +1,321 @@
...
+        metricOptions={{
+          metrics: slotRankings?.vals_largest_rewards,
+          metricsFmt: getSolStringWithFourDecimals,
+        }}
+      />
+    </>
+  );
+}
+
+interface MetricOptions<T extends number | bigint> {
```
atsai-jumptrading: nit but in case i was being confusing, i meant the whole props of like `icon, label, color...` being an interface would make the component read a bit cleaner

src/consts.ts
```diff
@@ -77,3 +77,12 @@ export const slotNavWidth =
...
 export const maxZIndex = 110;
+
+export const numQuickSearchSlots = 3;
+export const quickSearchCardWidth = 226;
+export const slotSearchGap = 40;
+export const slotSearchPadding = 20;
+export const slotSearchMaxWidth =
+  2 * slotSearchPadding +
+  numQuickSearchSlots * quickSearchCardWidth +
+  (numQuickSearchSlots - 1) * slotSearchGap;
```
atsai-jumptrading: i think we should just put all these in the component file, or make a consts file in that folder for colocation and we're not using these values anywhere else globally. Makes intellisense a bit cleaner too since it won't be giving as many suggestions to import stuff that only applies on one component

atsai-jumptrading (approved):

src/features/SlotDetails/SlotSearch.tsx
```diff
@@ -0,0 +1,321 @@
...
+          )
+        : undefined,
+    [firstProcessedLeaderIndex, leaderSlots],
+  );
+  const mostRecentQuickSlots = useMemo(
+    () =>
+      nextLeaderIndex !== undefined
+        ? leaderSlots
+            ?.slice(
+              Math.max(nextLeaderIndex - numQuickSearchSlots, 0),
```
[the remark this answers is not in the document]
amliu-jumptrading: Hmm, I guess for earliestQuickSlots I could slice it starting at firstProcessedLeaderIndex until its end and have the component chop it to numQuickSearchSlots. For mostRecentQuickSlots I can slice it until nextLeaderIndex and reverse that whole array. I thought it would be better to reverse a shorter list than one of n length, but the difference is probably negligible

PR #274, opened 2026-03-18
feat: add tooltips to txn execution duration charts

amliu-jumptrading:
- Add tooltips to txn execution duration charts
- Align y-axis of the cu and count charts
- sync tooltips on cu and count charts

### DIFF
--- a/src/features/SlotDetails/DetailedSlotStats/PerformanceSection/TxnExecutionDurationCharts/TxnExecutionDurationChartTooltip.tsx
+++ b/src/features/SlotDetails/DetailedSlotStats/PerformanceSection/TxnExecutionDurationCharts/TxnExecutionDurationChartTooltip.tsx
@@ -1,0 +1,41 @@
+import { Grid } from "@radix-ui/themes";
+import UplotTooltip from "../../../../../uplotReact/UplotTooltip";
+import { ChartTooltipRow } from "../../ChartTooltipRow";
+import { computeUnitsColor } from "../../../../../colors";
+import { txnExecutionDurationTooltipElId } from "../../../../Overview/SlotPerformance/ComputeUnitsCard/consts";
+
+interface TxnExecutionDurationChartTooltipProps {
+  data?: {
+    bucketIdx: number;
+    cuYVal: number | null | undefined;
+    countYVal: number | null | undefined;
+  };
+  formatBucketRange: (idx: number) => string;
+}
+
+export default function TxnExecutionDurationChartTooltip({
+  data,
+  formatBucketRange,
+}: TxnExecutionDurationChartTooltipProps) {
+  return (
+    <UplotTooltip elId={txnExecutionDurationTooltipElId}>
+      {data && (
+        <Grid columns="auto auto" gapX="2">
+          <ChartTooltipRow
+            label="Duration"
+            value={formatBucketRange(data.bucketIdx)}
+          />
+          <ChartTooltipRow
+            label="Avg CUs"
+            value={data.cuYVal ? Math.round(data.cuYVal).toLocaleString() : "0"}
+            color={computeUnitsColor}
+          />
+          <ChartTooltipRow
+            label="Count"
+            value={data.countYVal ? data.countYVal.toLocaleString() : "0"}
+          />
+        </Grid>
+      )}
+    </UplotTooltip>
+  );
+}
--- a/src/features/SlotDetails/DetailedSlotStats/PerformanceSection/TxnExecutionDurationCharts/index.tsx
+++ b/src/features/SlotDetails/DetailedSlotStats/PerformanceSection/TxnExecutionDurationCharts/index.tsx
@@ -1,20 +1,124 @@
 import { Box } from "@radix-ui/themes";
-import { useMemo } from "react";
+import { useCallback, useMemo, useState } from "react";
 import AutoSizer from "react-virtualized-auto-sizer";
 import uPlot, { type AlignedData, type Options } from "uplot";
 import UplotReact from "../../../../../uplotReact/UplotReact";
 import { useAtomValue } from "jotai";
 import { selectedSlotAtom } from "../../../../Overview/SlotPerformance/atoms";
 import { useSlotQueryResponseTransactions } from "../../../../../hooks/useSlotQuery";
-import type { SlotTransactions } from "../../../../../api/types";
-import { getMax, isDefined } from "../../../../../utils";
+import { getMax } from "../../../../../utils";
 import { getDurationWithUnits } from "../../../../Overview/SlotPerformance/TransactionBarsCard/chartUtils";
 import { SlotDetailsSubSection } from "../../SlotDetailsSubSection";
+import { chartAxisColor } from "../../../../../colors";
+import { txnExecutionDurationTooltipPlugin } from "./txnExecutionDurationTooltipPlugin";
+import { syncXScalePlugin } from "../../../../../uplotReact/syncXScalePlugin";
+import { syncYAxisPlugin } from "../../../../../uplotReact/syncYAxisPlugin";
+import { getAxisSize } from "../../../../../uplotReact/utils";
+import TxnExecutionDurationChartTooltip from "./TxnExecutionDurationChartTooltip";
+import { txnExecutionDurationScaleKey } from "../../../../Overview/SlotPerformance/ComputeUnitsCard/consts";
+
+const bucketCount = 20;
+const bucketIndices = Array.from({ length: bucketCount }, (_, i) => i);
+
+const getBucketIndex = (duration: number, maxDuration: number) =>
+  Math.trunc((duration / maxDuration) * bucketCount);
+
+const getBucketDuration = (idx: number, maxDuration: number) =>
+  (idx / bucketCount) * maxDuration;
+
+const axisFont = "8px Inter Tight";
+const barFillColor = "#3C2E69";
+const ySeriesIdx = 1;
+
+const barsPaths =
+  uPlot.paths?.bars?.({ size: [0.8] }) ??
+  (() => ({ stroke: new Path2D(), fill: new Path2D() })); // fallback no-op
+
+interface TxnExecutionDurationChartData {
+  cuData: AlignedData;
+  countData: AlignedData;
+  maxDuration: number;
+}
+
+const defaultTxnExecutionDurationChartData: TxnExecutionDurationChartData = {
+  cuData: [[], []],
+  countData: [[], []],
+  maxDuration: 0,
+};
+
+const fmtDuration = (nanos: number, maximumFractionDigits?: number) => {
+  const { value, unit } = getDurationWithUnits(
+    nanos,
+    false,
+    maximumFractionDigits,
+  );
+  return `${value} ${unit}`;
+};
 
 export default function TxnExecutionDurationCharts() {
   const selectedSlot = useAtomValue(selectedSlotAtom);
   const transactions =
     useSlotQueryResponseTransactions(selectedSlot).response?.transactions;
+  const [tooltipDataIdx, setTooltipDataIdx] = useState<number | undefined>();
+
+  const { cuData, countData, maxDuration }: TxnExecutionDurationChartData =
+    useMemo(() => {
+      if (!transactions) return defaultTxnExecutionDurationChartData;
+
+      const items = transactions.txn_landed
+        .map((_, i) => ({
+          duration: Number(
+            transactions.txn_mb_end_timestamps_nanos[i] -
+              transactions.txn_mb_start_timestamps_nanos[i],
+          ),
+          cu: transactions.txn_compute_units_consumed[i],
+        }))
+        .filter(({ duration }) => duration > 0);
+
+      if (items.length === 0) return defaultTxnExecutionDurationChartData;
+
+      const maxDuration = getMax(items.map(({ duration }) => duration)) + 1;
+
+      const buckets = Array.from({ length: bucketCount }, () => ({
+        count: 0,
+        cus: 0,
+      }));
+
+      for (const { duration, cu } of items) {
+        const idx = getBucketIndex(duration, maxDuration);
+        buckets[idx].count++;
+        buckets[idx].cus += cu;
+      }
+
+      const bucketedCounts = buckets.map(({ count }) => count);
+      const bucketedCUs = buckets.map(({ count, cus }) =>
+        count ? cus / count : 0,
+      );
+
+      return {
+        countData: [bucketIndices, bucketedCounts],
+        cuData: [bucketIndices, bucketedCUs],
+        maxDuration,
+      };
+    }, [transactions]);
+
+  const formatBucketRange = useCallback(
+    (idx: number) =>
+      `${fmtDuration(getBucketDuration(idx, maxDuration), 2)} - ${fmtDuration(getBucketDuration(idx + 1, maxDuration), 2)}`,
+    [maxDuration],
+  );
+
+  const tooltipData = useMemo(
+    () =>
+      tooltipDataIdx !== undefined
+        ? {
+            bucketIdx: tooltipDataIdx,
+            cuYVal: cuData[ySeriesIdx][tooltipDataIdx],
+            countYVal: countData[ySeriesIdx][tooltipDataIdx],
+          }
+        : undefined,
+    [countData, cuData, tooltipDataIdx],
+  );
 
   if (!transactions) return;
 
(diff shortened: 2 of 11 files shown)


src/features/SlotDetails/DetailedSlotStats/PerformanceSection/TxnExecutionDurationCharts/index.tsx
```diff
@@ -1,20 +1,149 @@
...
+        count: 0,
+        cus: 0,
+      }));
+      for (const { duration, cu } of items) {
+        const idx = getBucketIndex(duration, maxDuration);
+        buckets[idx].count++;
+        buckets[idx].cus += cu;
+      }
+
+      const indices = buckets.map((_, i) => i);
```
asuzuki-jumptrading: nit: we could create this outside the component since bucketCount is a const

src/features/SlotDetails/DetailedSlotStats/PerformanceSection/TxnExecutionDurationCharts/index.tsx
```diff
@@ -26,183 +155,186 @@ export default function TxnExecutionDurationCharts() {
...
+  const plotData = useMemo<AlignedData>(
+    () =>
+      log
+        ? [
+            data[0],
+            (data[ySeriesIdx] as number[]).map((v) => Math.log((v ?? 0) + 1)),
+          ]
+        : data,
+    [data, log],
+  );
```
asuzuki-jumptrading: Why not log the data in the parent where we're collecting the values?

src/features/SlotDetails/DetailedSlotStats/PerformanceSection/TxnExecutionDurationCharts/index.tsx
```diff
@@ -26,183 +155,186 @@ export default function TxnExecutionDurationCharts() {
...
+
+  const lastXIdx = Math.max(0, data[0].length - 1);
```
asuzuki-jumptrading: When would this be a problem? Isn't data[0].length going to be the bucketCount and never 0?

src/features/SlotDetails/DetailedSlotStats/PerformanceSection/TxnExecutionDurationCharts/index.tsx
```diff
@@ -26,183 +155,186 @@ export default function TxnExecutionDurationCharts() {
...
             );
-            if (longestVal !== "") {
-              self.ctx.font = axis.font?.[0] ?? "8px Inter Tight";
-              axisSize +=
-                self.ctx.measureText(longestVal).width / devicePixelRatio;
-            }
-            const px = Math.ceil(axisSize);
-            return px;
+
+            if (longestVal === "") return axisSpacing;
```
asuzuki-jumptrading: Is it ok that we got rid of Math.ceil for this value?

src/uplotReact/syncXScalePlugin.ts
```diff
@@ -19,7 +21,7 @@ export function syncXScalePlugin(): uPlot.Plugin {
         syncInProgress = true;
         let min = xScale.min ?? 0;
         let max = xScale.max ?? 0;
-        if (max - min < 100) {
+        if (minRange > 0 && max - min < minRange) {
```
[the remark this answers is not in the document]
amliu-jumptrading: Good points, I'll update the lines below too.

src/features/SlotDetails/DetailedSlotStats/PerformanceSection/TxnExecutionDurationCharts/TxnExecutionDurationChartTooltip.tsx
```diff
@@ -0,0 +1,45 @@
...
+  formatBucketRange,
+}: TxnExecutionDurationChartTooltipProps) {
+  return (
+    <UplotTooltip elId={elId}>
+      {data && (
+        <Grid columns="auto auto" gapX="2">
+          <ChartTooltipRow
+            label="Duration"
+            value={data.bucketIdx}
+            formatter={formatBucketRange}
```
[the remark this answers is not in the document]
amliu-jumptrading: That makes sense to me. Updated :)

src/features/SlotDetails/DetailedSlotStats/PerformanceSection/TxnExecutionDurationCharts/index.tsx
```diff
@@ -1,20 +1,149 @@
...
+        .filter(({ duration }) => duration > 0);
+
+      if (items.length === 0) return defaultTxnExecutionDurationChartData;
+
+      const maxDuration = getMax(items.map(({ duration }) => duration)) + 1;
+
+      const buckets = new Array(bucketCount).fill(0).map(() => ({
+        count: 0,
+        cus: 0,
+      }));
```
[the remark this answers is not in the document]
amliu-jumptrading: True, updated it but with bucketCount instead of slotsPerLeader here

src/features/SlotDetails/DetailedSlotStats/PerformanceSection/TxnExecutionDurationCharts/index.tsx
```diff
@@ -26,183 +155,186 @@ export default function TxnExecutionDurationCharts() {
...
             // bail out, force convergence
-            // eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-member-access
-            if (cycleNum > 1) return (axis as any)._size;
-            let axisSize = axis.ticks?.size ?? 0 + (axis.gap ?? 0);
-            // adding tick gap
-            axisSize += 5;
-            // find longest value
+            if (cycleNum > 1) return getAxisSize(axis);
+
+            const axisSpacing = (axis.ticks?.size ?? 0) + (axis.gap ?? 0);
```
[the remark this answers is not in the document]
amliu-jumptrading: I think that +5 was there because the axis size was being calculated incorrectly before and not using the axis.gap if axis.ticks.size existed (correct me if this is wrong). Now that the gap is always being used, we should not need to add an additional 5 tick gap. Manually checking the chart visuals, the axis looks good to me without adding an additional gap.

src/features/SlotDetails/DetailedSlotStats/PerformanceSection/TxnExecutionDurationCharts/index.tsx
amliu-jumptrading: I can add it back 👍

src/features/SlotDetails/DetailedSlotStats/PerformanceSection/TxnExecutionDurationCharts/index.tsx
```diff
@@ -26,183 +155,186 @@ export default function TxnExecutionDurationCharts() {
...
           size(self, values, axisIdx, cycleNum) {
             const axis = self.axes[axisIdx];
             // bail out, force convergence
-            // eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-member-access
-            if (cycleNum > 1) return (axis as any)._size;
-            let axisSize = axis.ticks?.size ?? 0 + (axis.gap ?? 0);
-            // adding tick gap
-            axisSize += 5;
-            // find longest value
+            if (cycleNum > 1) return getAxisSize(axis);
```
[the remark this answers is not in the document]
amliu-jumptrading: axis._size is always a number as far as I can see. I just added the nullish coalescing value of 0 for safety since we're using an uPlot internal variable not exposed through their api 😅

src/features/SlotDetails/DetailedSlotStats/PerformanceSection/TxnExecutionDurationCharts/index.tsx
amliu-jumptrading: We can't only store the logged data because we need the original values for the tooltip. We'd need to have another array loggedCountData in the parent anyways. In that case, I'd prefer to keep the logging logic inside this Chart component because the logging is specifically related to how we display the data on the chart

src/features/SlotDetails/DetailedSlotStats/PerformanceSection/TxnExecutionDurationCharts/index.tsx
amliu-jumptrading: Yes technically for our use case since we're only using the Chart component for cuData and countData, but I didn't want make assumptions about data[0].length in the component since it's a generic and can take in any `AlignedData`.

src/features/SlotDetails/DetailedSlotStats/PerformanceSection/TxnExecutionDurationCharts/index.tsx
```diff
@@ -1,20 +1,149 @@
...
+    cu.redraw(false, true);
+    count.redraw(false, true);
+
+    const cuSize: number = getAxisSize(cu.axes[ySeriesIdx]);
+    const countSize: number = getAxisSize(count.axes[ySeriesIdx]);
+    minYAxisSizeRef.current = Math.max(cuSize, countSize);
+
+    cu.redraw(false, true);
+    count.redraw(false, true);
+  }, [transactions, uplotCharts]);
```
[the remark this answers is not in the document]
amliu-jumptrading: While trying to move this logic into a plugin, I realized that although this would work for the initial chart, the axis widths also change when zooming. Triggering an onSizeChanged from ready (which only fires once on chart fully init) wouldn't be able to handle this case. To be clear, my original solution also doesn't handle that case correctly either.
The approach I landed upon instead handles the syncing of the y axis by wrapping each chart's size function, recording their natural sizes, and batching the redraws in a requestAnimationFrame. This also consolidates the logic that I was passing into Chart originally with the minYAxisSizeRef.

src/features/SlotDetails/DetailedSlotStats/PerformanceSection/TxnExecutionDurationCharts/txnExecutionDurationTooltipPlugin.ts
```diff
@@ -0,0 +1,17 @@
+import type uPlot from "uplot";
+import { baseTooltipPlugin } from "../../../../../uplotReact/baseTooltipPlugin";
+import { txnExecutionDurationTooltipElId } from "../../../../Overview/SlotPerformance/ComputeUnitsCard/consts";
+
+export function txnExecutionDurationTooltipPlugin(
+  setTooltipDataIdx: (idx: number) => void,
+): uPlot.Plugin {
+  function showOnCursor(_u: uPlot, _xVal: number, idx: number | null) {
+    if (idx === null) return false;
```
asuzuki-jumptrading: I think you can remove the `null` idx case after the last PR update

src/features/SlotDetails/DetailedSlotStats/PerformanceSection/TxnExecutionDurationCharts/index.tsx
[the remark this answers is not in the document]
asuzuki-jumptrading: Ah yes, thanks!

src/uplotReact/syncYAxisPlugin.ts
```diff
@@ -0,0 +1,92 @@
...
+          });
+        } else {
+          group.charts.set(chartId, u);
+        }
+      },
+
+      destroy: (u) => {
+        const group = syncGroups.get(syncKey);
+        if (!group) return;
+        if (group.charts.get(chartId) !== u) return;
```
asuzuki-jumptrading, line 74: When would this happen?

src/uplotReact/syncYAxisPlugin.ts
```diff
@@ -0,0 +1,92 @@
...
+  const group = syncGroups.get(key);
+  if (!group || group.pendingSync !== null) return;
+
+  const newMaxSize = Math.max(0, ...group.naturalSizes.values());
+  if (newMaxSize === group.maxSize) return;
+
+  group.maxSize = newMaxSize;
+  group.pendingSync = requestAnimationFrame(() => {
+    for (const chart of group.charts.values()) {
+      chart.redraw(false, true);
```
asuzuki-jumptrading, line 23: This redraw would get triggered on every draw. Could we check that the natural size differs from the max size to trigger the redraw?

src/uplotReact/syncYAxisPlugin.ts
```diff
@@ -0,0 +1,92 @@
...
+  return {
+    opts: (_u, chartOpts) => {
+      const yAxis = chartOpts.axes?.[syncAxisIdx];
+      if (!yAxis) return;
+
+      const originalYAxisSize = yAxis.size;
+
+      yAxis.size = (self, values, axisIdx, cycleNum) => {
+        const original =
+          typeof originalYAxisSize === "function"
```
asuzuki-jumptrading, line 44: Just curious -- is originalYAxisSize always a function when the chart is zoomed?

src/uplotReact/syncYAxisPlugin.ts
amliu-jumptrading: We're storing the charts by chartId because I noticed that on chart load, 1 destroy followed by 2 ready callbacks were being called. I also observed this for our other plugins so this may be a quirk of uplot/plugins. Due to destroy being called first though, a stale chart would get added to group.charts when ready was called the first time and it would never get cleaned up properly. By storing them by chartId, we'd be able to overwrite the charts to use the latest one for that chartId.

This introduced a race condition we need to guard against here though: 2 ready callbacks are called followed by 1 destroy would cause the destroy corresponding to the initial ready to clear the new chart instead of acting as a no-op.

src/uplotReact/syncYAxisPlugin.ts
amliu-jumptrading: No, size is a function or number depending on what gets passed into chart options. Zoomed is dictated by scale, not size. When zoomed, the y-scale's range function will produce a [min, max] which uplot generates new tick values and splits from. Those values are passed into the y-axis's size function if it is a function.

src/uplotReact/syncYAxisPlugin.ts
asuzuki-jumptrading: Discussed offline -- the double trigger is likely due to React strict mode. The race condition is probably a uPlot issue.

src/uplotReact/syncYAxisPlugin.ts
amliu-jumptrading: The `if (newMaxSize === group.maxSize) return;` will guard against redraws when the group's maxSize has not changed, so it will only trigger on draws that result in a new max y axis sizes (e.g. initial draw, zoom). We can't just check that the natural size differs from the max size for redraws because of the following case:
```
chart A is 100px, chart B is 50px. max size is 100px
the user zooms and chart A is 75px, chart B is 25px. max size is 75px
we want to redraw both chart A and chart B to use an axis size of 75px (if we skip the redraw for chart A, it will still have an axis size of 100px)
```

asuzuki-jumptrading (dismissed):

src/uplotReact/syncXScalePlugin.ts
```diff
@@ -5,7 +5,9 @@ import { getDefaultStore } from "jotai";
 const store = getDefaultStore();
 let syncInProgress = false;
 
-export function syncXScalePlugin(): uPlot.Plugin {
+export function syncXScalePlugin(opts?: { minRange?: number }): uPlot.Plugin {
+  const minRange = opts?.minRange ?? 100;
```
atsai-jumptrading: could make 0 the default, bit more generic, since it is kinda arbitrary to be 100 which just happens to make sense for the prior implementations

src/features/SlotDetails/DetailedSlotStats/PerformanceSection/TxnExecutionDurationCharts/index.tsx
```diff
@@ -26,183 +130,162 @@ export default function TxnExecutionDurationCharts() {
...
+        duration: { time: false },
+        y: {
+          auto: true,
+          /* When zoomed into empty buckets, the range can be an unreasonably
+           * large value on the log scale or go negative on the linear scale.
+           * Clamp the range to the data's max value or a reasonable number.
+           * The axis splits will be [0, 1, 2] with no data. */
+          range: (_u: uPlot, _min: number, dataMax: number) =>
+            log
+              ? [0, Math.max(dataMax, Math.log(3))]
```
atsai-jumptrading, line 199: why specifically log 3?

atsai-jumptrading (dismissed):

src/features/SlotDetails/DetailedSlotStats/PerformanceSection/TxnExecutionDurationCharts/index.tsx
amliu-jumptrading: When we're storing the plot data, we're calculating it as `Math.log(v + 1)`. When we convert the plot data to get the original value, we're calculating it as `Math.round(Math.exp(v) - 1)`.

When there is no data within range (or the values are very small), I want to show at least [0, 1, 2] (in the splits you can see we're using scaleMin, minpoint, and scaleMax). To guarantee that the max tick displays as 2 , we need the range max to be a value that satisfies `Math.exp(v) - 1 = 2`. `Math.log(3)` is the value that satisfies that equation.

atsai-jumptrading (approved):

PR #276, opened 2026-03-20
feat: animated status card slot number

asuzuki-jumptrading:
- animate status card slot number
- accept class name in AnimatedInteger component
- refactor card stat sizing options

### DIFF
--- a/src/components/AnimatedInteger.tsx
+++ b/src/components/AnimatedInteger.tsx
@@ -6,34 +6,25 @@ import {
   useRef,
   useState,
   type CSSProperties,
-  type ElementRef,
 } from "react";
 import styles from "./animatedInteger.module.css";
 import clsx from "clsx";
-import {
-  Box,
-  Flex,
-  Text,
-  type FlexProps,
-  type TextProps,
-} from "@radix-ui/themes";
+import { Box, Flex, Text, type FlexProps } from "@radix-ui/themes";
 import { useUnmount } from "react-use";
 import useIsDocumentVisible from "../hooks/useIsDocumentVisible";
 
 const MIN_ANIMATION_DURATION_MS = 20;
-const WINDOW_HEIGHT_CSS_VAR = "--number-window-height";
 
 interface AnimatedIntegerProps {
   value: number;
   animationDurationMs?: number;
   /**
-   * Explicitly set height of text container. If omitted, calculate the default text height
+   * Explicitly set height of text container. If omitted, uses default text height.
    */
   height?: number;
-  textSize?: TextProps["size"];
   containerRowJustify?: FlexProps["justify"];
+  className?: string;
 }
-
 /**
  * Animate individual digits to reach the desired value.
  * Animation occurs one step at a time, even if the value jumps by > 1.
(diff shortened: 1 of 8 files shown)


src/components/CardStat.tsx
```diff
@@ -1,43 +1,53 @@
...
+    >
       <Text className={styles.label}>{label}</Text>
       <Flex align="baseline" gap="1">
-        <Text
-          className={styles.value}
-          style={{ color: valueColor, ...valueStyles }}
-        >
-          {value}
-        </Text>
+        {isNumber(value) && animateInteger ? (
```
amliu-jumptrading: nit: `typeof value === "number"` should be equivalent to `isNumber(value)` here without using lodash. I'm only commenting on this since we've mentioned wanting to remove lodash to decrease bundle size eventually.

src/components/CardStat.tsx
```diff
@@ -1,43 +1,53 @@
...
+  animateInteger?: boolean;
 }
-
 export default function CardStat({
   label,
   value,
   valueColor,
   appendValue,
-  large,
+  valueSize = "small",
```
amliu-jumptrading: nit: It seems like "small" and "medium" are the most common cases for valueSize in the app (and they are about equally frequent). Since there is no clear majority, I have a slight preference to remove the default and make valueSize required so it's immediately visible at the component callsite which size is being used (at the cost of slightly more verbosity).

amliu-jumptrading (dismissed):

src/components/CardStat.tsx
asuzuki-jumptrading: makes sense, will do :) thanks!

amliu-jumptrading (dismissed):

src/components/AnimatedInteger.tsx
```diff
@@ -23,17 +24,24 @@ import useIsDocumentVisible from "../hooks/useIsDocumentVisible";
 const MIN_ANIMATION_DURATION_MS = 20;
 const WINDOW_HEIGHT_CSS_VAR = "--number-window-height";
 
+interface TextStyles {
+  size?: TextProps["size"];
```
atsai-jumptrading: i don't see size ever being used

src/components/AnimatedInteger.tsx
```diff
@@ -47,28 +55,30 @@ export default function AnimatedInteger(props: AnimatedIntegerProps) {
...
-    // re-measure height on text size change
+    // re-measure height on text styling change
     setMeasuredHeight(undefined);
-  }, [props.height, props.textSize]);
+  }, [props.height, props.textStyles?.size, props.textStyles?.className]);
 
   useLayoutEffect(() => {
     if (height != null || measureHeightRef.current == null) return;
+    // measure height before paint
     setMeasuredHeight(measureHeightRef.current.getBoundingClientRect().height);
```
atsai-jumptrading: could you remove all the height measuring and just use 1lh instead? Since you're just using the height in a css calc later. Could still override with the height prop. Not a big performance increase but would simplify the component? And wouldn't need to make a css var with a specific height

src/components/AnimatedInteger.tsx
asuzuki-jumptrading: ah yeah we don't need it anymore... will remove

src/components/AnimatedInteger.tsx
asuzuki-jumptrading: oh good point! I'll try this out, thanks!

src/components/AnimatedInteger.tsx
atsai-jumptrading: can just pass the classname then instead of an object

src/components/AnimatedInteger.tsx
asuzuki-jumptrading: I was able to refactor this to not need manual height measuring or 1lh :D
We make the parent grow to fit the selectable span. And then the animated part (a sibling to selectable span) has position: absolute and fits in the parent

src/components/AnimatedInteger.tsx
```diff
@@ -231,18 +195,25 @@ function AnimatedIntegerInner({
...
+    <Flex position="relative" justify={containerRowJustify}>
+      {/* Add hidden Text to set container height and to support selection / copy of the full number (digits components are not selectable). */}
+      <Text
+        className={clsx(className, styles.selectionText)}
+        style={heightStyle}
+      >
+        {currentValidNumber}
+      </Text>
+
+      <Flex height="100%" position="absolute" inert="true">
```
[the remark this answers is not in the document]
asuzuki-jumptrading: This is incorrect. Absolutely positioned children of flex containers behave as if they are the sole flex item in the flex container ([noted here](https://www.w3.org/TR/css-flexbox-1/#abspos-items)), so this element will overlay the selection text correctly.

amliu-jumptrading (approved):

atsai-jumptrading (approved):

PR #291, opened 2026-04-03
Performance slop airstrike

ripatel-fd:
I asked Claude to make the website faster

I have no idea why, but this reduces CPU usage from 110% to 70%
on Mac M4 Pro.

I think it's a mix of Bars.tsx using div instead of svg,
and view culling

I also tried to reduce now() calls but it didn't really work

And some weird React state management tricks IDK

🦧


ripatel-fd:
Closing this but leaving the optimizations up for future reference

PR #298, opened 2026-04-13
chore: performance improvement for Bars

asuzuki-jumptrading:
Adapted from @ripatel-fd https://github.com/firedancer-io/firedancer-frontend/pull/291

- Instead of using SVGs for Bars, use a div with pseudo elements, background color and masking
- Remove `useMeasure` which was triggering forced reflow. Use CSS vars and calc to determine widths instead

### DIFF
--- a/src/features/StartupProgress/Firedancer/bars.module.css
+++ b/src/features/StartupProgress/Firedancer/bars.module.css
@@ -1,34 +1,109 @@
 .bars {
-  width: 100%;
+  --container-width: 100%;
+
+  position: relative;
+  width: var(--container-width);
   height: var(--bar-height, 50px);
 
-  rect {
-    fill: var(--boot-progress-gossip-bars-color);
-    &.threshold {
-      fill: var(--app-teal);
-    }
-    &.filled {
-      fill: var(--boot-progress-gossip-filled-bar-color);
-    }
+  --step: calc(var(--bar-width) + var(--bar-gap));
+  /*
+    n bars, n - 1 gaps
+    n = floor((container_width + gap) / step)
+    used_width = n * step - gap 
+    = floor((container_width + gap) / step) * step - gap
+    = round(down, container_width + gap, step) - gap
+    NOTE: floor(x / y) * y = round(down, x, y)
+  */
+  --used-width: max(
+    0px,
+    calc(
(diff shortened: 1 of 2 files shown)


src/features/StartupProgress/Firedancer/bars.module.css
```diff
@@ -1,34 +1,107 @@
...
+  /*
+    n bars, n - 1 gaps
+    n = foor((width + gap) / step)
+    usedwidth = n * step - gap 
+    = foor((width + gap) / step) * step - gap
+    = round(down, width + gap, step) - gap
+    NOTE: floor(x / y) * y = round(down, x, y)
+  */
+  --used-width: calc(
+    round(down, 100% + var(--bar-gap), var(--step)) - var(--bar-gap)
```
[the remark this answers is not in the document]
asuzuki-jumptrading: Unlikely to happen, given step is usually around 5px. But will add for safety

atsai-jumptrading (dismissed):

src/features/StartupProgress/Firedancer/bars.module.css
```diff
@@ -1,34 +1,104 @@
 .bars {
+  position: relative;
   width: 100%;
```
amliu-jumptrading: nit: should we add a css variable `--container-width: 100%` and use it for the width here and the calculation for `--used-width`?

src/features/StartupProgress/Firedancer/bars.module.css
```diff
@@ -1,34 +1,104 @@
...
-    &.threshold {
-      fill: var(--app-teal);
-    }
-    &.filled {
-      fill: var(--boot-progress-gossip-filled-bar-color);
-    }
+  --step: calc(var(--bar-width) + var(--bar-gap));
+  /*
+    n bars, n - 1 gaps
+    n = floor((width + gap) / step)
```
amliu-jumptrading: nit: can we refer to this width as `containerWidth` to distinguish from the bar width? It makes it a bit clearer since we're referring to bar gap when we use gap, but container width when we use width.

amliu-jumptrading (dismissed):

amliu-jumptrading (approved):

PR #305, opened 2026-04-20
Remove page faults and scheduler info from tile metrics

ripatel-fd:


amliu-jumptrading:
We've moved the columns into their own group at the rightmost end to de-emphasize them
https://github.com/firedancer-io/firedancer-frontend/pull/319

PR #316, opened 2026-05-11
chore: adjust supermajority table contrast and default expand accordions

amliu-jumptrading:
- adjust text color for better contrast
  - offline should be more prominent than online because that is the data we want to draw attention to
- truncate version text at end, not start
- default open both accordions

### DIFF
--- a/src/features/StartupProgress/Firedancer/Supermajority/SupermajorityTable.tsx
+++ b/src/features/StartupProgress/Firedancer/Supermajority/SupermajorityTable.tsx
@@ -49,7 +49,7 @@ export default function SupermajorityTable({
   const size: Size = isXNarrow ? "xnarrow" : isNarrow ? "narrow" : "wide";
 
   const [isOfflineExpanded, setIsOfflineExpanded] = useState(true);
-  const [isOnlineExpanded, setIsOnlineExpanded] = useState(false);
+  const [isOnlineExpanded, setIsOnlineExpanded] = useState(true);
 
   const supermajorityEpoch = useAtomValue(supermajorityEpochAtom);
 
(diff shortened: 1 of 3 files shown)


src/features/StartupProgress/Firedancer/Supermajority/supermajorityTable.module.css
```diff
@@ -94,10 +102,8 @@
 
 .peer {
   column-gap: 4px;
-  color: var(--gray-11);
-  &.offline {
-    color: var(--gray-9);
 
+  &:not(.offline) {
```
asuzuki-jumptrading: nit: could we flip the logic and add the class `.online` instead, so we can remove the `&:not` checks?

src/features/StartupProgress/Firedancer/Supermajority/SupermajorityTable.tsx
```diff
@@ -395,7 +395,7 @@ function ClientVersionCell({
       size={size}
       className={clsx(styles.version, { [styles.offline]: isOffline })}
     >
-      <Flex width="35px">
+      <Flex minWidth="35px">
```
asuzuki-jumptrading: What does this change?

asuzuki-jumptrading (dismissed):

src/features/StartupProgress/Firedancer/Supermajority/SupermajorityTable.tsx
amliu-jumptrading: With just width, the flex was shrinking when the version text is long, so I added a minWidth. After inspecting the element a bit more, I noticed that the non-placeholder double icon are 36.66px instead of 35px. I'll change this to `<Flex width="37px" flexShrink="0">` to also handle the case of the flex container contents being larger.

src/features/StartupProgress/Firedancer/Supermajority/supermajorityTable.module.css
amliu-jumptrading: Sure! I also noticed that if I re-arrange some of the css styles, I don't need to pass isOffline into some of the cell components, so alongside adding `.online`, I also updated some of the other styles

asuzuki-jumptrading (approved):

atsai-jumptrading (approved):

PR #320, opened 2026-05-11
feat: filter by client id on leader schedule; wait for peers to load …

asuzuki-jumptrading:
…before filtering

- fix race condition where filtering happened before peers were loaded, so results showed no peers matches
- allow filtering by client ID
- remove `;` for multi-value filtering (use commas instead). Added redirect to be backwards compatible for previous links with `;`

### DIFF
--- a/src/atoms.ts
+++ b/src/atoms.ts
@@ -551,21 +551,32 @@ export const myStakePctAtom = atom((get) => {
   return (Number(stake) / Number(totalActivePeersStake)) * 100;
 });
 
-export const allLeaderNamesAtom = atom((get) => {
+export const leaderScheduleSearchDependenciesAtom = atom((get) => {
   const epoch = get(epochAtom);
   const peers = get(peersAtom);
+  const peersCount = get(peersCountAtom);
 
-  if (!epoch || !peers) return;
+  if (!epoch || peersCount === 0) return;
+  return {
+    epoch,
+    peers,
+  };
+});
+
+export const allLeaderNamesClientIdsAtom = atom((get) => {
+  const dependencies = get(leaderScheduleSearchDependenciesAtom);
+  if (!dependencies) return;
+
+  const { epoch, peers } = dependencies;
 
   const uniquePubkeys = new Set(
     epoch.leader_slots.map((i) => epoch.staked_pubkeys[i]),
   );
-  const leadersWithNames = [...uniquePubkeys].map((pubkey) => ({
+  return [...uniquePubkeys].map((pubkey) => ({
     pubkey: pubkey,
     name: peers[pubkey]?.info?.name?.toLowerCase(),
+    clientId: peers[pubkey]?.gossip?.client_id,
   }));
-
-  return leadersWithNames;
 });
 
 type SupermajorityEpochByPubkey = Map<
--- a/src/features/LeaderSchedule/Search.tsx
+++ b/src/features/LeaderSchedule/Search.tsx
@@ -8,11 +8,10 @@ import {
   Reset,
   Tooltip,
 } from "@radix-ui/themes";
-import { atom, useAtomValue, useSetAtom } from "jotai";
+import { useAtomValue, useSetAtom } from "jotai";
 import { setSearchLeaderSlotsAtom, searchLeaderSlotsAtom } from "./atoms";
 import { useMount } from "react-use";
 import {
-  currentLeaderSlotAtom,
   discountedLateVoteSlotsAtom,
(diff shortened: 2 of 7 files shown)


src/features/LeaderSchedule/Search.tsx
```diff
@@ -69,13 +65,11 @@ export default function Search() {
...
   });
 
-  if (!isVisible) return;
-
   return (
     <Flex className={styles.container} gap="2" wrap="wrap">
       <Box className={styles.searchBox}>
         <TextField.Root
-          placeholder="Name, pubkey, or slot (separate with , or ; for multiple values)"
+          placeholder="Name, Client Id, Pubkey, or Slot (use , for multiple)"
```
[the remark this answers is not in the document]
asuzuki-jumptrading: This is the name we're using

src/features/LeaderSchedule/atoms.ts
```diff
@@ -43,15 +43,28 @@ export const setSearchLeaderSlotsAtom = atom(
...
-      ?.filter(({ name, pubkey }) =>
+    const matchingClientIds = Object.entries(clientIdToClientName).reduce(
+      (acc, [clientId, clientName]) => {
+        if (searchTextParts.some((s) => clientName.toLowerCase().includes(s))) {
+          acc.add(parseInt(clientId));
+        }
+        return acc;
+      },
+      new Set<number>(),
+    );
```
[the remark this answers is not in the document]
asuzuki-jumptrading: This is fine as is

src/features/LeaderSchedule/atoms.ts
```diff
@@ -43,15 +43,28 @@ export const setSearchLeaderSlotsAtom = atom(
...
+
+    const leaders = get(allLeaderNamesClientIdsAtom);
+    const searchPubkeys = leaders
+      ?.filter(({ name, pubkey, clientId }) =>
         searchTextParts.some(
```
amliu-jumptrading: Can we do `(clientId != null && matchingClientIds.has(clientId)) || searchTextParts.some(name?.includes(s) || pubkey.toLowerCase().includes(s))` since the clientId check doesn't use `s`

src/features/LeaderSchedule/index.tsx
```diff
@@ -1,12 +1,22 @@
...
+import {
+  currentLeaderSlotAtom,
+  leaderScheduleSearchDependenciesLoadedAtom,
+  slotNavFilterAtom,
+  slotOverrideAtom,
+} from "../../atoms";
 import { useMount } from "react-use";
 import { clusterIndicatorHeight, headerHeight } from "../../consts";
 
+const isVisibleAtom = atom((get) => !!get(currentLeaderSlotAtom));
```
amliu-jumptrading: Can we check for `!!get(currentLeaderSlotAtom) && !!get(leaderScheduleSearchDependenciesAtom)` here and get rid of the `leaderScheduleSearchDependenciesLoadedAtom` atom?

src/features/LeaderSchedule/atoms.ts
asuzuki-jumptrading: oh yes that makes more sense!

src/features/LeaderSchedule/index.tsx
asuzuki-jumptrading: yes, that's better :) thanks!

amliu-jumptrading (dismissed):

src/atoms.ts
```diff
@@ -551,21 +541,31 @@ export const myStakePctAtom = atom((get) => {
   return (Number(stake) / Number(totalActivePeersStake)) * 100;
 });
 
-export const allLeaderNamesAtom = atom((get) => {
+export const leaderScheduleSearchDependenciesAtom = atom((get) => {
   const epoch = get(epochAtom);
   const peers = get(peersAtom);
 
-  if (!epoch || !peers) return;
+  if (!epoch || !peers || Object.keys(peers).length === 0) return;
```
atsai-jumptrading: could use PeersListAtom? Then we don't need to iterate it again each time

atsai-jumptrading (dismissed):

src/atoms.ts
asuzuki-jumptrading: Oh true! I could even use `peersCountAtom` :D

atsai-jumptrading: even better!

atsai-jumptrading (approved):

amliu-jumptrading (approved):

PR #321, opened 2026-05-12
feat: repair server metrics

drubin-fd:


src/features/Overview/LiveNetworkMetrics/consts.ts
```diff
@@ -29,6 +31,7 @@ export const networkMaxByteValues: {
     gossip: 1_000_000_000 / 8,
     tpu: 1_000_000 / 8,
     repair: 1_000_000 / 8,
+    rserve: 1_000_000 / 8,
```
amliu-jumptrading: You may want to set this to a larger value such as `100_000_000` 😄 The bar currently looks always maxed out because values are hovering around `~5_000_000-12_000_000`. If you have a more accurate max byte value, you can also use that instead.

drubin-fd: Yep, thank you. I had no idea going in what value would make sense here. I've landed on 100m in 25m out being a good number.

amliu-jumptrading (approved):

PR #324, opened 2026-05-12
feat: supermajority table tumbling numbers count

amliu-jumptrading:
- animate total node count for online/offline accordions with tumbling numbers animation
- add added/removed node count w/ 5m window

### DIFF
--- a/src/atoms.ts
+++ b/src/atoms.ts
@@ -595,22 +595,87 @@ export const supermajorityEpochAtom = (function getSupermajorityEpochAtom() {
   );
 })();
 
+type DeltaEvent = {
+  addPeers: string[];
+  removePeers: string[];
+  timestamp: number;
+};
+
+export const SUPERMAJORITY_DELTA_WINDOW_MS = 5 * 60_000;
+
 export const [
   supermajorityOnlinePeersAtom,
   updateSupermajorityOnlinePeersAtom,
-] = (function getSupermajorityOnlinePeersAtoms() {
+  supermajorityCountsAtom,
+  deleteSupermajorityDeltaEntriesAtom,
+  resetSupermajorityAtom,
+] = (function getSupermajorityAtoms() {
   const _supermajorityOnlinePeersAtom = atomWithImmer(new Set<string>());
+  const _supermajorityInitialized = atom(false);
+  const _supermajorityWindowStartAtom = atomWithImmer(new Set<string>());
+  const _supermajorityHistoryAtom = atomWithImmer<DeltaEvent[]>([]);
+
   return [
     atom((get) => get(_supermajorityOnlinePeersAtom)),
-    atom(null, (_get, set, addPeers: string[], removePeers: string[]) => {
+    atom(null, (get, set, addPeers: string[], removePeers: string[]) => {
       set(_supermajorityOnlinePeersAtom, (draft) => {
-        for (const peer of addPeers) {
-          draft.add(peer);
-        }
-        for (const peer of removePeers) {
-          draft.delete(peer);
+        for (const peer of addPeers) draft.add(peer);
+        for (const peer of removePeers) draft.delete(peer);
+      });
+
+      // When recording history, ignore the first update because it will
+      // contain initial state, not recent additions/removals
+      if (!get(_supermajorityInitialized)) {
+        set(_supermajorityInitialized, true);
+        set(_supermajorityWindowStartAtom, (draft) => {
+          for (const peer of addPeers) draft.add(peer);
+          for (const peer of removePeers) draft.delete(peer);
+        });
+      } else if (addPeers.length > 0 || removePeers.length > 0) {
+        const timestamp = Date.now();
+        const event: DeltaEvent = { addPeers, removePeers, timestamp };
+        set(_supermajorityHistoryAtom, (draft) => {
+          draft.push(event);
+        });
+      }
+    }),
+    atom((get) => {
+      const windowStartOnline = get(_supermajorityWindowStartAtom);
+      const currentOnline = get(_supermajorityOnlinePeersAtom);
+
+      let offline = 0;
+      for (const peer of windowStartOnline) {
+        if (!currentOnline.has(peer)) offline++;
+      }
+      const online = currentOnline.size + offline - windowStartOnline.size;
+
+      return { online, offline };
+    }),
+    atom(null, (get, set) => {
+      const history = get(_supermajorityHistoryAtom);
+      if (history.length === 0) return;
+
+      const cutoff = Date.now() - SUPERMAJORITY_DELTA_WINDOW_MS;
+      const cutoffIdx = history.findIndex((event) => event.timestamp >= cutoff);
+      if (cutoffIdx === 0) return;
+
+      const numExpiredEvents = cutoffIdx === -1 ? history.length : cutoffIdx;
+      set(_supermajorityWindowStartAtom, (draft) => {
+        for (let i = 0; i < numExpiredEvents; i++) {
+          const { addPeers, removePeers } = history[i];
+          for (const peer of addPeers) draft.add(peer);
+          for (const peer of removePeers) draft.delete(peer);
         }
       });
+      set(_supermajorityHistoryAtom, (draft) =>
+        cutoffIdx === -1 ? [] : draft.slice(cutoffIdx),
+      );
+    }),
+    atom(null, (_get, set) => {
+      set(_supermajorityOnlinePeersAtom, new Set());
+      set(_supermajorityInitialized, false);
+      set(_supermajorityWindowStartAtom, new Set());
+      set(_supermajorityHistoryAtom, []);
     }),
   ];
 })();
--- a/src/features/StartupProgress/Firedancer/Supermajority/SupermajorityCounts.tsx
+++ b/src/features/StartupProgress/Firedancer/Supermajority/SupermajorityCounts.tsx
@@ -1,0 +1,28 @@
+import { Flex, Text } from "@radix-ui/themes";
+import styles from "./supermajorityCounts.module.css";
+import {
+  SUPERMAJORITY_DELTA_WINDOW_MS,
+  supermajorityCountsAtom,
+} from "../../../../atoms";
+import { useAtomValue } from "jotai";
+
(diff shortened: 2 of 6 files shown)


src/features/StartupProgress/Firedancer/Supermajority/useCountDelta.ts
```diff
@@ -0,0 +1,64 @@
+import { useEffect, useRef, useState } from "react";
+
+const DELTA_WINDOW_MS = 5_000;
+export const DELTA_WINDOW_LABEL = `Last ${DELTA_WINDOW_MS / 1_000}s`;
+const SAMPLE_INTERVAL_MS = 250;
+
```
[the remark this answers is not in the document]
amliu-jumptrading: I updated the sample interval to 1000ms

src/features/StartupProgress/Firedancer/Supermajority/CountDelta.tsx
```diff
@@ -0,0 +1,17 @@
...
+  pubkeys: Set<string>;
+}
+export function CountDelta({ pubkeys }: CountDeltaProps) {
+  const { added, removed } = useCountDelta(pubkeys);
+
+  return (
+    <Flex className={styles.container} align="center" gap="5px">
+      <Text>{DELTA_WINDOW_LABEL}</Text>
+      <Text className={styles.added}>+{added}</Text>
+      <Text className={styles.removed}>-{removed}</Text>
```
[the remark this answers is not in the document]
amliu-jumptrading: This is intentional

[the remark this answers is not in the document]
amliu-jumptrading: No, the colors are correct. added should always be green and removed should always be removed regardless of if they are moving to online/offline.

asuzuki-jumptrading (dismissed):

src/features/StartupProgress/Firedancer/Supermajority/countUtils.ts
```diff
@@ -0,0 +1,179 @@
+export const SUPERMAJORITY_DELTA_WINDOW_MS = 5 * 60_000;
+export const SUPERMAJORITY_DELTA_WINDOW_LABEL = `Last ${SUPERMAJORITY_DELTA_WINDOW_MS / 60_000}m`;
```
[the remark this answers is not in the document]
amliu-jumptrading: The window was increased from 5s to 5m based on user feedback

src/features/StartupProgress/Firedancer/Supermajority/countUtils.ts
```diff
@@ -0,0 +1,179 @@
...
+
+export function pushDeltaEvent(state: PeersState, event: DeltaEvent) {
+  const { initialState, currentState, eventsCount } = state;
+
+  for (const pubkey of event.online) {
+    eventsCount.set(pubkey, (eventsCount.get(pubkey) ?? 0) + 1);
+
+    const initialStatus = initialState.get(pubkey);
+    if (initialStatus === undefined) {
+      // infer that a new peer coming online was initially offline
```
asuzuki-jumptrading: Is it possible to set the initial states on the first call to updateSupermajorityOnlinePeersAtom?

src/features/StartupProgress/Firedancer/Supermajority/countUtils.ts
```diff
@@ -0,0 +1,179 @@
...
+      state.online++;
+      continue;
+    }
+
+    const previousStatus = currentState.get(pubkey);
+    if (
+      previousStatus === undefined ||
+      previousStatus === PubkeyStatus.Online
+    ) {
+      // no-op if the new status is the same as the existing status
```
asuzuki-jumptrading: When would this happen?

src/features/StartupProgress/Firedancer/Supermajority/SupermajorityCounts.tsx
```diff
@@ -0,0 +1,24 @@
...
+  const counts = useAtomValue(supermajorityCountsAtom);
+
+  return (
+    <Flex className={styles.container} align="center" gap="5px">
+      <Text>{SUPERMAJORITY_DELTA_WINDOW_LABEL}</Text>
+      <Text className={styles.added}>
+        +{isOffline ? counts.offline : counts.online}
+      </Text>
+      <Text className={styles.removed}>
+        -{isOffline ? counts.online : counts.offline}
```
asuzuki-jumptrading, line 24: Should the `-` be inside the `{}` so you don't end up with a double negative text?

src/features/StartupProgress/Firedancer/Supermajority/countUtils.ts
amliu-jumptrading: This is an edge case where two different events have the pubkey with the same status (with no events in between that contain that pubkey).

event 0:
- online: [pkA]
- offline: []

event n:
- online: [pkA]
- offline: []

When event 0 expires, the previous status of pkA is online and the current status is also online because of event n. In practice, this shouldn't happen since an already online validator should not go online again without having gone offline in between, but data-wise it's possible so I've added a guard.

The previousStatus === undefined check is to satisfy the type checker

amliu-jumptrading: After chatting with the backend, this case would be a bug on the backend so I'm going to remove handling for it.

src/features/StartupProgress/Firedancer/Supermajority/countUtils.ts
amliu-jumptrading: The initial state here refers to the state of the pubkey at the beginning of the rolling 5m window, not necessarily when the user initially loads the page (when the first call to updateSupermajorityOnlinePeersAtom happens).  Currently, I'm removing the pubkey from initialState and currentState if they are not referenced by any event during the rolling 5m window, which is where we'd hit this case of not having the initialState of the pubkey stored. Instead of clearing the initialState and currentState though, I can keep a track of the rolling initial and current states (setting them initially on the first call to updateSupermajorityOnlinePeersAtom). After chatting with the backend, it's also not possible for an unknown pubkey (not recorded in the initial update) to be added during supermajority (which was another edge case I had this code block handling).

src/features/StartupProgress/Firedancer/Supermajority/SupermajorityCounts.tsx
amliu-jumptrading: The counts should never be negative so we won't hit a case of double negative text.

src/features/StartupProgress/Firedancer/Supermajority/__tests__/countUtils.test.ts
```diff
@@ -37,7 +32,7 @@ describe("pushDeltaEvent", () => {
   });
 
   it("counts peers going offline", () => {
-    const state = defaultPeersState();
+    const state = getPeerState(new Set(["pkA", "pkB", "pkC"]));
```
amliu-jumptrading: With the implementation change, the tests must make sure the initial peer state includes any pubkeys that are intended to go from online to offline (otherwise any offline events will be interpreted as an offline to offline no-op)

src/features/StartupProgress/Firedancer/Supermajority/__tests__/countUtils.test.ts
```diff
@@ -89,28 +80,8 @@ describe("pushDeltaEvent", () => {
     expect(counts(state)).toEqual({ online: 2, offline: 1 });
   });
 
-  it("consecutive offline or online events for the same peer are ignored", () => {
```
amliu-jumptrading: Consecutive offline or consecutive online events are not valid and would be considered a bug in the validator, so I'm removing testing for this case in the UI.

src/atoms.ts
```diff
@@ -609,11 +606,22 @@ export const supermajorityEpochAtom = (function getSupermajorityEpochAtom() {
...
-] = (function getSupermajorityOnlinePeersAtoms() {
+  supermajorityCountsAtom,
+  deleteSupermajorityDeltaEntriesAtom,
+  resetSupermajorityAtom,
+] = (function getSupermajorityAtoms() {
   const _supermajorityOnlinePeersAtom = atomWithImmer(new Set<string>());
+  const _supermajorityInitialized = atom(false);
+  const _supermajorityDeltaAtom = atomWithImmer<{
+    history: DeltaEvent[];
+    state: PeersState;
```
asuzuki-jumptrading: Could we rely on `_supermajorityOnlinePeersAtom` instead of storing `currentOnline`, `offline` and `online` in `PeerState`?
I think the `_supermajorityDeltaAtom` could just keep track of the events array and the state at the start of the events array.
The `supermajorityCountsAtom` could be a derived atom that compares `_supermajorityOnlinePeersAtom` and `_supermajorityDeltaAtom`'s start state.

amliu-jumptrading: Sounds good, I'm going to separate the `_supermajorityDeltaAtom` into a `_supermajorityHistoryAtom` and `_supermajorityWindowStartAtom` since there is less state that needs to be grouped together.

src/atoms.ts
```diff
@@ -606,22 +595,88 @@ export const supermajorityEpochAtom = (function getSupermajorityEpochAtom() {
...
+    atom((get) => {
+      const windowStartOnline = get(_supermajorityWindowStartAtom);
+      const currentOnline = get(_supermajorityOnlinePeersAtom);
+      let online = 0;
+      let offline = 0;
+
+      for (const peer of windowStartOnline)
+        if (!currentOnline.has(peer)) offline++;
+      for (const peer of currentOnline)
+        if (!windowStartOnline.has(peer)) online++;
```
asuzuki-jumptrading: Could we do `online = currentOnline.size - (windowStartOnline.size - offline)` so we don't need to iterate over currentOnline?

asuzuki-jumptrading (dismissed):
Nice!

asuzuki-jumptrading (dismissed):

atsai-jumptrading (dismissed):

asuzuki-jumptrading (dismissed):

asuzuki-jumptrading (approved):

PR #330, opened 2026-05-14
chore: prevent overscroll on startup overlay

amliu-jumptrading:
macOS has a feature commonly called "rubber band scrolling" or "elastic scrolling" where scrolling past the edge of the page content causes a bounce-back effect. This has UX benefits of providing visual feedback when the user has reached the edge of the content, but for our startup overlay, it reveals the app underneath and looks unnatural. Adding `overscroll-behavior: none` here prevents this behavior for the startup overlay only.


asuzuki-jumptrading (approved):
🎉

PR #337, opened 2026-05-19
chore: web worker shreds cache

asuzuki-jumptrading:
- move shreds websocket message handling to web worker


src/api/worker/cache/shredsCalc.ts
```diff
@@ -0,0 +1,281 @@
...
+            max: data.range.max,
+          }
+        : undefined;
+  }
+
+  function setRecursiveDeleteTimeout() {
+    setTimeout(
+      () => {
+        deleteSlots(
+          !!validatorStateCache.isStartup,
```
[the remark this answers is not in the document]
asuzuki-jumptrading: fixed with stop function

src/api/worker/cache/shreds/__tests__/shredsCalc.test.ts
```diff
@@ -0,0 +1,576 @@
...
+    // before subscribe
+    expect(liveShredsCache.get(subKey)).toBeUndefined();
+
+    liveShredsCache.subscribe(subKey, {
+      publishIntervalMs: Infinity,
+    });
+
+    // initial state after subscribe
+    expect(liveShredsCache.get(subKey)).toEqual({});
+
```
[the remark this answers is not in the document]
asuzuki-jumptrading: fixed with afterEach() fix

src/api/worker/cache/shredsCalc.ts
```diff
@@ -0,0 +1,289 @@
...
+            data.slotsShreds.referenceTs,
+          )
+        ) {
+          // once we find a slot that is complete and far enough in the past,
+          // delete all slot numbers less it but keep this one for label spacing reference
+          shouldDeleteSlot = true;
+          continue;
+        }
+
+        if (shouldDeleteSlot) {
```
amliu-jumptrading: Can we move this check to line 149 (beginning of the loop) and add a continue to the end of this if block? We would be able to remove the `!shouldDeleteSlot` guard on line 153 and skip any checks for any of the shouldDeleteSlot steps.

Also it would make sure to delete any slots that have a null/undefined maxEventTsDelta instead of skipping them.

src/api/worker/cache/shredsCache.ts
```diff
@@ -0,0 +1,62 @@
+import type { LiveShredsItem, ValidatorStateCache } from "../types";
+import {
+  createBatchPublisher,
+  type PublisherEntry,
+  type PublisherOptions,
+} from "./batchPublisher";
+import type { LiveShreds } from "../../types";
+import {
+  createShredsCalc as createShredsCalc,
```
amliu-jumptrading: nit: ✂️

src/api/worker/messageHandler.ts
```diff
@@ -68,10 +76,35 @@ export function createMessageHandler(post: (msg: FromWorkerMessage) => void) {
     post({ type: "historyArray", items }),
   );
 
+  const validatorStateCache: ValidatorStateCache = {};
```
amliu-jumptrading: Can we make it so `ValidatorStateCache` requires `isStartup` to be defined and initialize it as false here so we can be explicit about the default and don't have to coerce the field into a boolean when it's used?

src/api/worker/cache/__tests__/shredsCalc.test.ts
```diff
@@ -0,0 +1,575 @@
...
+    });
+
+    // add initial shreds
+    liveShredsCache.add(subKey, {
+      reference_slot: 0,
+      reference_ts: 0n,
+      slot_delta: Object.values(events).map((v) => v.slot),
+      shred_idx: Object.values(events).map((v) => 0),
+      event: Object.values(events).map((v) => v.e),
+      event_ts_delta: Object.values(events).map((v) => v.ts),
```
amliu-jumptrading: `events` is already an array so we can just use `events.map` here

src/api/worker/cache/shredsCalc.ts
```diff
@@ -0,0 +1,289 @@
...
+  delayMs,
+} from "../../../features/Overview/ShredsProgression/const";
+import type { ValidatorStateCache } from "../types";
+import type {
+  LiveShredsData,
+  ShredEventTsDeltaMs,
+  ShredEventTsDeltas,
+  Slot,
+  SlotsShreds,
+} from "../../../features/Overview/ShredsProgression/types";
```
amliu-jumptrading: The worker owns the shreds data pipeline, so it would be preferable to have the consts and types defined in this file or in files under the worker directory (`shredConsts.ts` and `shredTypes.ts`) and the UI components that use them can consume them. It makes it more clear that they originate from the worker's shred calculation and keeps the worker more self-contained.

src/api/worker/types.ts
```diff
@@ -9,6 +9,7 @@ import {
   supermajoritySchema,
 } from "../entities";
 import type { GossipHealthEma } from "../atoms";
+import type { LiveShredsData } from "../../features/Overview/ShredsProgression/types";
```
amliu-jumptrading: Same here on my comment about moving the shred types somewhere into the worker directly. It will keep the worker more self contained.

src/api/worker/cache/shreds/shredsCalc.ts
```diff
@@ -0,0 +1,289 @@
...
+        : POST_STARTUP_DELETE_INTERVAL_MS,
+    );
+  }
+
+  function stop() {
+    clearTimeout(timeoutId);
+    timeoutId = undefined;
+  }
+
+  setRecursiveDeleteTimeout();
```
amliu-jumptrading: Instead of starting the delete timeout when the shred calc is created, can we start it when the first event is added (`if (timeoutId == null) setRecursiveDeleteTimeout()`) and call `stop()` when `reset` is called. This way the deletion check is not running when it doesn't need to be.

src/api/worker/messageHandler.ts
```diff
@@ -68,10 +76,35 @@ export function createMessageHandler(post: (msg: FromWorkerMessage) => void) {
...
+
+  const liveShredsCache = createLiveShredsCache(
```
amliu-jumptrading: You should pass a type in to restrict what keys can be used to subscribe to liveShredsCache. `createLiveShredsCache<LiveShredKey>`
`LiveShredKey` can be defined in `src/api/worker/types` alongside the other keys

Right now the other subscribe/update calls in `messageHandler` will throw type errors if the wrong key is used, but the liveShredsCache one won't

amliu-jumptrading: You can also update the `LiveShredsItem` type to use `LiveShredKey` as the type for key so that `updateLiveShredsObject` is type aware of all the possible values for `key` and you can remove the default case from the switch

src/api/worker/messageHandler.ts
asuzuki-jumptrading: I'd prefer to leave this state as undefined when we don't know the state.
Shreds cache is the only place using this field, and will treat missing info as not startup. But there may be other caches in the future that use this data and want to handle missing startup info differently

src/api/worker/cache/shreds/shredsCalc.ts
asuzuki-jumptrading: Discussed offline -- deletion is light if there is no data. The time when the cache exists but no data is stored is minimal (during disconnect or during startup before catching up phase), so it's ok not to stop and restart the deletion timeout

src/api/worker/messageHandler.ts
asuzuki-jumptrading: Discussed offline -- shreds cache will only ever use one key at a time, so we can move the key to a const within the shreds cache

src/api/worker/cache/shreds/shredsCalc.ts
```diff
@@ -0,0 +1,283 @@
...
+
+export const xRangeMs = 10_000;
+export const delayMs = 50;
+export const STARTUP_DELETE_INTERVAL_MS = 1_000;
+export const POST_STARTUP_DELETE_INTERVAL_MS = xRangeMs / 4;
+
+export function createShredsCalc(getValidatorState: () => ValidatorState) {
+  let data: LiveShredsData = {};
+
+  function add({
```
asuzuki-jumptrading, line 22: logic from https://github.com/firedancer-io/firedancer-frontend/blob/dc78fa04483d8daa2af6b3f68c6024830b8b4ee7/src/features/Overview/ShredsProgression/atoms.ts#L84

src/api/worker/cache/shreds/shredsCalc.ts
```diff
@@ -0,0 +1,283 @@
...
+
+  function reset() {
+    data = {};
+  }
+
+  /**
+   * Delete slots that completed before the chart x-axis starting time, or with dots outside visible x range
+   * Update the min slot
+   */
+  function deleteSlots(isStartup: boolean, serverTimeNanos?: number) {
```
asuzuki-jumptrading, line 105: logic copied from https://github.com/firedancer-io/firedancer-frontend/blob/dc78fa04483d8daa2af6b3f68c6024830b8b4ee7/src/features/Overview/ShredsProgression/atoms.ts#L163

src/api/worker/cache/shreds/__tests__/shredsCalc.test.ts
```diff
@@ -0,0 +1,575 @@
...
+} from "../shredsCalc";
+
+function getDefaultValidatorState() {
+  return {
+    isStartup: undefined,
+    serverTimeNanos: undefined,
+  };
+}
+
+describe("createShredsCalc", () => {
```
asuzuki-jumptrading: logic copied from https://github.com/firedancer-io/firedancer-frontend/blob/dc78fa04483d8daa2af6b3f68c6024830b8b4ee7/src/features/Overview/ShredsProgression/__tests__/atoms.test.tsx#L15

src/api/worker/cache/shreds/shredsCache.ts
```diff
@@ -0,0 +1,74 @@
...
+        key,
+        subscribed: false,
+        lastPublishMs: 0,
+        publishIntervalMs: options.publishIntervalMs,
+        shredsCalc: createShredsCalc(getValidatorState),
+      };
+    },
+    collect: (e) => {
+      return { key: e.key, data: e.shredsCalc.data };
+    },
```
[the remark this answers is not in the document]
asuzuki-jumptrading: This will be mitigated by unsubscribing on disconnect. While connected, we expect to receive updates every 50ms (which is publishIntervalMs) so we wouldn't expect unnecessary unchanged data sends.
FYI the next steps (not in this PR) are to have individual shreds charts subscribe to receive shreds data only when rendered.

src/api/worker/messageHandler.ts
```diff
@@ -68,10 +76,40 @@ export function createMessageHandler(post: (msg: FromWorkerMessage) => void) {
...
 
+      if (isEntry(item, "summary", "server_time_nanos")) {
+        validatorState.serverTimeNanos = item.value;
+      }
+      if (
+        isEntry(item, "summary", "startup_progress") ||
+        isEntry(item, "summary", "boot_progress")
+      ) {
+        validatorState.isStartup = item.value.phase !== "running";
+      }
```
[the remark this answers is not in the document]
asuzuki-jumptrading: fixed

src/api/worker/messageHandler.ts
```diff
@@ -68,10 +76,40 @@ export function createMessageHandler(post: (msg: FromWorkerMessage) => void) {
...
   return {
+    onConnectionChange(msg: {
+      type: "connected" | "connecting" | "disconnected";
+    }): void {
+      if (msg.type !== "connected") {
+        validatorState = defaultValidatorState;
+        liveShredsCache.reset();
+      }
+      post(msg);
+    },
```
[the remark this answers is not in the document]
asuzuki-jumptrading: fixed with an unsubscribe

src/features/Overview/ShredsProgression/atoms.ts
```diff
@@ -1,330 +1,35 @@
...
 
-export type SlotsShreds = {
-  referenceTs: number;
-  // slot number to Slot
-  slots: Map<number, Slot>;
-};
+export const liveShredsPostStartupRangeAtom = atom((get) => {
+  const range = get(liveShredsDataAtom)?.range;
+  const slotCaughtUp = get(slotCaughtUpAtom);
+  if (!range || slotCaughtUp == null) return;
```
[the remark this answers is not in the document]
asuzuki-jumptrading: this is not necessary; the type is already inferred to be `LiveShredsData | undefined`

src/api/worker/cache/shreds/shredsCalc.ts
asuzuki-jumptrading: After looking at this some more, I've changed my mind and will start / stop the timer on add and reset as you've suggested. I'll also limit the shredsCache functions to make sure timer cleanup gets fired

src/api/worker/messageHandler.ts
```diff
@@ -68,10 +76,47 @@ export function createMessageHandler(post: (msg: FromWorkerMessage) => void) {
...
   return {
+    onConnectionChange(msg: {
+      type: "connected" | "connecting" | "disconnected";
+    }): void {
+      if (msg.type !== "connected") {
+        validatorState = { ...defaultValidatorState };
+        liveShredsCache.unsubscribe();
+        liveShredsCache.reset();
+      }
+      post(msg);
```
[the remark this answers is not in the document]
asuzuki-jumptrading: updated with timer cleanup

src/api/worker/cache/shreds/shredsCache.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  return {
+    ...publisher,
+
+    reset() {
+      return publisher.reset(liveShredsKey);
+    },
+
+    unsubscribe() {
+      return publisher.unsubscribe(liveShredsKey);
```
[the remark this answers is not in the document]
asuzuki-jumptrading: updated reset function

src/api/useSetAtomWsData.ts
```diff
@@ -169,6 +169,18 @@ export function useSetAtomWsData() {
...
+          setLiveShreds(data);
+          break;
+      }
+    },
+    [setLiveShreds],
+  );
+
   const onMessage = useCallback(
     (msg: FromWorkerMessage) => {
       switch (msg.type) {
```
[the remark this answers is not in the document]
asuzuki-jumptrading: updated with publishNow function

src/api/worker/messageHandler.ts
```diff
@@ -68,10 +76,47 @@ export function createMessageHandler(post: (msg: FromWorkerMessage) => void) {
...
+    liveShredsOptions,
+    (items) => post({ type: "liveShredsObject", items }),
+    getValidatorState,
+  );
+
   return {
+    onConnectionChange(msg: {
+      type: "connected" | "connecting" | "disconnected";
+    }): void {
+      if (msg.type !== "connected") {
```
amliu-jumptrading, line 95: Should we be checking for `msg.type === "disconnected"` here?

Previously we were using `isSocketDisconnected` to trigger deleting live shreds data, but here we'll be doing so on both `"connecting"` and `"disconnected"`. Do we need to reset data during the `"connecting"` state?
`const isSocketDisconnected = useAtomValue(socketStateAtom) === SocketState.Disconnected;`

amliu-jumptrading (dismissed):
Good job on the refactor!

src/api/worker/messageHandler.ts
asuzuki-jumptrading: Thanks for checking! Yes, it's ok to just clear the state as soon as we're not "connected", because we would be replacing all of the shreds data after connecting again anyway.

src/api/worker/cache/shreds/shredsCalc.ts
```diff
@@ -0,0 +1,284 @@
...
+        }
+      }
+    }
+
+    // update range to reflect remaining slots
+    const remainingSlotNumbers = data.slotsShreds.slots.keys();
+    data.range =
+      data.range && data.slotsShreds.slots.size
+        ? {
+            min: Math.min(...remainingSlotNumbers),
```
atsai-jumptrading: can this get too big? Math.min might error out if too many remaining slot numbers?

src/api/worker/cache/shreds/shredsCalc.ts
```diff
@@ -0,0 +1,284 @@
...
+          }
+        : undefined;
+  }
+
+  function setRecursiveDeleteTimeout() {
+    deleteTimeoutId = setTimeout(
+      () => {
+        const validatorState = getValidatorState();
+        deleteSlots(!!validatorState.isStartup, validatorState.serverTimeNanos);
+        setRecursiveDeleteTimeout();
```
atsai-jumptrading: we might need some error handling in case deleteSlots errors out and the recursion just stops

src/api/worker/cache/shreds/shredsCalc.ts
```diff
@@ -0,0 +1,284 @@
...
+        for (
+          let slotNumber = minSlot;
+          slotNumber <= data.range.max - 50;
+          slotNumber++
+        ) {
+          const slot = data.slotsShreds.slots.get(slotNumber);
+          if (!slot) continue;
+          data.slotsShreds.slots.delete(slotNumber);
+        }
+        minSlot = data.range.max - 50;
```
atsai-jumptrading:
```suggestion
        minSlot = data.range.max - 50 + 1;
```
I think we're off by 1 here? minSlot is a deleted slot here

atsai-jumptrading (dismissed):

src/api/worker/cache/shreds/shredsCalc.ts
asuzuki-jumptrading: good point... will add

src/api/worker/cache/shreds/shredsCalc.ts
asuzuki-jumptrading: Very unlikely, but possible. I'll use a loop instead

src/api/worker/cache/shreds/shredsCalc.ts
asuzuki-jumptrading: Good catch! I'll fix this. Fortunately, it didn't impact the visualization because we have guards against missing slots in a range.

atsai-jumptrading (approved):

PR #339, opened 2026-05-20
fix: rename/reorder txn states

jherrera-jump:
In anticipation of the corresponding backend changes:

https://github.com/firedancer-io/firedancer/pull/9869

- maps the new names to the old ones to stay forwards/backwards compatible. This will keep the slot details page most functional on older backends.
- In firedancer, load actually comes before validate, so we need to swap these. This is technically a breaking change, but it should only cause the load/validate states to display incorrectly which isn't a dealbreaker.

Will merge the backend support alongside a build with this commit.

### DIFF
--- a/src/api/entities.ts
+++ b/src/api/entities.ts
@@ -378,31 +378,51 @@ export const bootProgressSchema = z.object({
   catching_up_first_replay_slot: z.number().nullable().optional(),
 });
 
-export const slotTransactionsSchema = z.object({
-  start_timestamp_nanos: z.coerce.bigint(),
-  target_end_timestamp_nanos: z.coerce.bigint(),
-  txn_mb_start_timestamps_nanos: z.coerce.bigint().array(),
-  txn_mb_end_timestamps_nanos: z.coerce.bigint().array(),
-  txn_compute_units_requested: z.number().array(),
-  txn_compute_units_consumed: z.number().array(),
-  txn_transaction_fee: z.coerce.bigint().array(),
-  txn_priority_fee: z.coerce.bigint().array(),
-  txn_tips: z.coerce.bigint().array(),
-  txn_error_code: z.number().array(),
-  txn_from_bundle: z.boolean().array(),
-  txn_is_simple_vote: z.boolean().array(),
-  txn_bank_idx: z.number().array(),
-  txn_preload_end_timestamps_nanos: z.coerce.bigint().array(),
-  txn_start_timestamps_nanos: z.coerce.bigint().array(),
-  txn_load_end_timestamps_nanos: z.coerce.bigint().array(),
-  txn_end_timestamps_nanos: z.coerce.bigint().array(),
-  txn_arrival_timestamps_nanos: z.coerce.bigint().array(),
-  txn_microblock_id: z.number().array(),
-  txn_landed: z.boolean().array(),
-  txn_signature: z.string().array(),
-  txn_source_ipv4: z.string().array(),
-  txn_source_tpu: z.string().array(),
-});
+export const slotTransactionsSchema = z.preprocess(
+  (data) => {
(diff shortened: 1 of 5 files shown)


src/features/SlotDetails/DetailedSlotStats/ComputeSection/CumulativeExecutionTimeStats.tsx
```diff
@@ -104,14 +108,14 @@ export default function CumulativeExecutionTimeStats() {
           max={max}
         />
         <Row
-          label="Validating"
+          label={stateDisplayNames[TxnState.VALIDATE]}
```
asuzuki-jumptrading: This changes the label from "Validating" to "Validate" for Frankendancer. Is that intentional?

asuzuki-jumptrading (dismissed):

src/features/SlotDetails/DetailedSlotStats/ComputeSection/CumulativeExecutionTimeStats.tsx
jherrera-jump: not intentional but not an issue imo

src/api/entities.ts
```diff
@@ -378,31 +378,51 @@ export const bootProgressSchema = z.object({
...
+    txn_tips: z.coerce.bigint().array(),
+    txn_error_code: z.number().array(),
+    txn_from_bundle: z.boolean().array(),
+    txn_is_simple_vote: z.boolean().array(),
+    txn_bank_idx: z.number().array(),
+    txn_preload_end_timestamps_nanos: z.coerce.bigint().array(),
+    txn_start_timestamps_nanos: z.coerce.bigint().array(),
+    txn_load_end_timestamps_nanos: z.coerce.bigint().array(),
+    txn_end_timestamps_nanos: z.coerce.bigint().array(),
+    txn_commit_end_timestamps_nanos: z.coerce.bigint().array().optional(),
```
amliu-jumptrading, line 417: This looks like a new field added. Is this a new Firedancer-only field? Do we need to add a TxnState phase to show it?

src/features/SlotDetails/DetailedSlotStats/ComputeSection/ExecutionTime.tsx
```diff
@@ -133,30 +133,50 @@ function getX(pct: number) {
 }
 
 function Row({ label, value, max, minValue, maxValue }: RowProps) {
-  const pct = (value / max) * 100;
-  const minPct = (minValue / max) * 100;
-  const maxPct = (maxValue / max) * 100;
+  const hasData = isFinite(value) && isFinite(minValue) && isFinite(maxValue);
```
amliu-jumptrading, line 136: Why is it not guaranteed that we have data here anymore? Did a field become optional? It seems like the values would only be non-finite if `transactions.txn_mb_start_timestamps_nanos.length === 0`

asuzuki-jumptrading (approved):

src/api/entities.ts
jherrera-jump: this will be supported by both clients. We won't add a state but it might let us simplify the bundle logic a bit hopefully since we can use it instead of txn_mb_end, which is actually not quite what we want to be using.

src/features/SlotDetails/DetailedSlotStats/ComputeSection/ExecutionTime.tsx
jherrera-jump: an empty block (0 transactions) showed some weird Infinity / Nan / Nan thing here.

On testnet we get blocks without non-vote transacstions and no bundles pretty frequently

amliu-jumptrading (approved):

PR #342, opened 2026-05-28
chore: refactor shreds plugin to use xRange

asuzuki-jumptrading:
- Replace `tsXValueOffset` with `xRange`, which keeps track of what ts delta values and positions the x axis represents
- Helper function to draw startup axes

### DIFF
--- a/src/features/Overview/ShredsProgression/shredsProgressionPlugin.ts
+++ b/src/features/Overview/ShredsProgression/shredsProgressionPlugin.ts
@@ -45,31 +45,55 @@ export type LabelPositions = {
   };
 };
 
+type XRange = {
+  minDeltaTs: number;
+  maxDeltaTs: number;
+  minCanvasPos: number;
+  maxCanvasPos: number;
+  minCssPos: number;
+  maxCssPos: number;
+};
+
 export function shredsProgressionPlugin(
   isOnStartupScreen: boolean,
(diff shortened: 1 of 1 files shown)


src/features/Overview/ShredsProgression/shredsProgressionPlugin.ts
```diff
@@ -45,31 +45,55 @@ export type LabelPositions = {
...
+    // If we get ws messages buffered and it results in a temporary high
+    // diff, shred still move smoothly by using the avg
+    const now = Date.now();
+    const serverTimeMs = store.get(serverTimeMsAtom);
+
+    if (serverTimeMs) {
+      const timeDiff = now - serverTimeMs;
+      prevTimeDiffs.push(timeDiff);
+      while (prevTimeDiffs.length > 20) {
+        prevTimeDiffs.shift();
```
atsai-jumptrading, line 79: could change to a ring buffer

atsai-jumptrading (approved):

src/features/Overview/ShredsProgression/shredsProgressionPlugin.ts
asuzuki-jumptrading: True! I'm hoping to move this to an ema calc on the web worker later, or we could use the ring buffer that's there.

PR #345, opened 2026-06-01
chore: show program cache misses

amliu-jumptrading:
- show hits and misses instead of hits and lookups
- highlight changed text contiguously (every char to the right of the first changed char) instead of changed chars only

### DIFF
--- a/src/features/Overview/ProgramCacheCard/HitRateStat.tsx
+++ b/src/features/Overview/ProgramCacheCard/HitRateStat.tsx
@@ -21,9 +21,8 @@ type Status = "Unknown" | "Good" | "Average" | "Bad";
 
 type HitRateValues = {
   percentage: string;
-  numerator: string;
-  denominator: string;
-  showFraction: boolean;
+  hits: string;
+  misses: string;
   status: Status;
 };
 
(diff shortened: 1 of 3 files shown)


src/features/Overview/ProgramCacheCard/HitRateStat.tsx
```diff
@@ -79,10 +71,9 @@ export default function HitRateStat() {
             : "Good";
 
     return {
-      percentage: percentage === 100 ? "100" : percentage.toFixed(20),
-      numerator: hits.toLocaleString(),
-      denominator: lookups.toLocaleString(),
-      showFraction: Boolean(lookups),
+      percentage: (Math.trunc(percentage * 100) / 100).toFixed(2),
```
asuzuki-jumptrading, line 74: Why do we need to truncate before .toFixed?

asuzuki-jumptrading (approved):

src/features/Overview/ProgramCacheCard/HitRateStat.tsx
amliu-jumptrading: toFixed's behavior is to round then format to a fixed decimal precision. In this case, we want the truncated decimal (e.g. if the hit rate is 99.999 it should show 99.99 instead of 100.00 to ensure we're drawing attention to any cache misses) and the toFixed is there to ensure the value always has two decimals to prevent shifting/visual gaps (e.g. 100 as 100.00 and 99.9 as 99.90)

asuzuki-jumptrading: Ah yeah that makes sense. Thanks!

atsai-jumptrading (approved):

PR #354, opened 2026-06-15
feat: let users turn off the changing-number animation

theobolo:
The digit-tumbling count-up animation plays on every numeric update
(slot counters tick a few times per second), which some users find
distracting and which can briefly lag the displayed value behind the
real one. Add a "Animate changing numbers" toggle under a new Display
settings popover in the header, persisted to localStorage, so each user
can opt out. The animation is also skipped when the OS requests reduced
motion. Default stays on, preserving the current experience.

### DIFF
--- a/src/components/AnimatedInteger.tsx
+++ b/src/components/AnimatedInteger.tsx
@@ -10,8 +10,10 @@ import {
 import styles from "./animatedInteger.module.css";
 import clsx from "clsx";
 import { Box, Flex, Text, type FlexProps } from "@radix-ui/themes";
-import { useUnmount } from "react-use";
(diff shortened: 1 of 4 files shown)


amliu-jumptrading:
Hi Theo, we’ve judged it as premature to add user settings to disable the Animated Numbers in the application. We added animations to provide a sense of progression through the epoch and show the validator is chugging along properly. Without it, the Slots Timeline at the top of the overview feels a bit static since the distance between the storage/root/voted/confirmed/processed/repair/turbine slots rarely changes even as they all progress forwards. We’ll keep an eye out on user sentiment though and consider adding user settings in the future.

PR #355, opened 2026-06-15
fix: view slot details regardless of current identity

theobolo:
## Summary

Fix slot details visibility for slots produced by this node under a previous identity, and improve handling of unavailable historical per-transaction data.

## Context

The slot details page previously gated validity on the current identity's leader slots. As a result, any slot not led by the current identity was shown as belonging to another validator.

This caused an issue after identity changes, for example after an unstake: slots actually produced by this node under a previous identity were hidden, because the server also recomputes the slot's `mine` flag against the new identity and therefore cannot always be used to recognize older slots correctly.

### Slot details validity

- Gate slot detail validity on data availability rather than the current identity's leader schedule.
- Query the requested slot's publish data before deciding whether the slot is valid.
- Treat any slot for which the server returns data as viewable.
- Keep the header resolution based on the original leader from the epoch schedule.

This resolves the previous chicken-and-egg issue where a slot was only considered valid once fetched, but was only fetched once considered valid.

### Detailed view rendering

- Render the detailed view as soon as any detailed field is available.
- Let each subsection hide itself when its own data is missing.
- Avoid gating the entire page on retained per-transaction data.
- Show a clear `not available` message instead of a perpetual spinner when historical per-transaction data is no longer retained.

This applies to:

- `DetailedSlotStats`
- Compute card
- Banks card

### Quick-search behavior

Update the Earliest / Most Recent quick-search logic so it reflects slots produced by this node, instead of slots from the current identity's leader schedule.

Produced slots are now identified using the server-stamped `mine` flag, unioned from:

- cached slot responses
- mine rankings

This is more accurate after an identity move or unstake. The leader schedule lists every slot for an identity even after it moves to another node, which can incorrectly surface slots produced elsewhere. The server-stamped `mine` flag is a better signal for what this node actually produced.

## Review notes

Parts of this change were AI-assisted.

The non-obvious areas worth reviewing carefully are:

- `mineProcessedSlotsAtom` in `src/atoms.ts`
  - Recovers which slots this node produced using the server `mine` flag.
  - Designed to remain robust across identity switches.

- `getSlotState` in `src/features/Overview/SlotPerformance/atoms.ts`
  - Implements the data-availability gate for slot detail visibility.

## Validation

Verified live against a Firedancer `v0.911` validator.

### DIFF
--- a/src/atoms.ts
+++ b/src/atoms.ts
@@ -7,6 +7,7 @@ import {
   identityKeyAtom,
   serverTimeNanosAtom,
   skippedSlotsAtom,
+  slotRankingsAtom,
   startupProgressAtom,
 } from "./api/atoms";
 import type {
@@ -228,6 +229,16 @@ export const slotPublishAtomFamily = atomFamily((slot?: number) =>
 export const slotResponseAtomFamily = atomFamily((slot?: number) =>
   atom((get) => (slot !== undefined ? get(slotResponseAtom)[slot] : undefined)),
 );
+
+/** Cached slots this node produced, per the server-stamped `mine` flag. */
+export const cachedMineSlotsAtom = atom((get) => {
+  const responses = get(slotResponseAtom);
+  const slots: number[] = [];
+  for (const key in responses) {
+    if (responses[key].publish.mine) slots.push(Number(key));
+  }
+  return slots;
+});
 
 export const setSlotResponseAtom = atom(
   null,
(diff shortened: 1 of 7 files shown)


amliu-jumptrading:
Hi Theo, I appreciate your contribution. It’s definitely a bug that the user is not able to see slots previously processed by this validator because its current pubkey is different. There are a couple of issues with this implementation though.

First, this PR bases slot validity on whether there is a slotPublish response for the given slot (`get(slotPublishAtomFamily(slot)) !== undefined`). Even if the slot is not mine, the server will send a response, making `get(slotPublishAtomFamily(slot)) !== undefined` always true when the slot is fetched before the validity check (also added by this PR). Instead, it would be more accurate to check `get(slotPublishAtomFamily(slot))?.mine`. The issue with this though is we periodically clear slot publish data that is not within our cache range (https://github.com/firedancer-io/firedancer-frontend/blob/main/src/atoms.ts#L300) to prevent memory build up, so it’s not guaranteed that slot publish data will stay available. The better approach is to request produced slots (or data that can be used to derive them) from the server. This would require a new server endpoint.

Another issue with the approach this PR takes is the `cachedMineSlotsAtom`. This atom is derived from `slotResponseAtom`, which has the same issue I described above where data is periodically cleared. Also, there will only be items available in the `slotResponseAtom` if data was previously requested within the current session. There will be no data available on a fresh page load and the user would not be able to see slots from the previous identity in the slot search.

We’ve discussed this bug as a team and will be adding a web socket message to derive all produced slots. I’ll work on a frontend fix that allows users to view slot details for any produced slots as well as categorizes My Slots across the app (slot details, leader schedule, etc.) as produced slots instead of slots associated to the current identity.

amliu-jumptrading:
This feature is still blocked on changes that need to be added to the server, so I'm closing this PR until those changes can be made.

PR #356, opened 2026-06-15
feat: merge Overview Status card into the Slots timeline

theobolo:
The Overview "Status" card largely duplicated the "Slots" timeline:
both showed the current slot, time-until-leader (with progress), and the
next leader slot. The only unique datum was the vote status.

Remove the Status card and surface the vote status as a compact badge in
the Slots timeline header, giving a more compact Overview without losing
information. The card grid drops from four to three columns at the widest
breakpoint accordingly.


amliu-jumptrading:
Hi Theo, I hear your point about duplicate information in the Overview cards. The data in the Status card is available in other parts of the Overview. Even the vote status, actually, is reflected in the Vote Health icon as part of the Health Pane in the Header. We do want to preserve the prominence of the current slot though even if it is technically available as the Processed Slot in the Slot Timeline. I’ll have a change out soon that reflects the layout we’re envisioning as well as additional adjustments for different screen sizes.

amliu-jumptrading:
This has been resolved with #396

PR #366, opened 2026-06-26
chore: optimize slots list renderer

asuzuki-jumptrading:
- Prevent re-renders by memoizing parts that don't need udpates

### DIFF
--- a/src/features/Navigation/SlotsRenderer.tsx
+++ b/src/features/Navigation/SlotsRenderer.tsx
@@ -1,12 +1,5 @@
-import { atom, useAtomValue } from "jotai";
-import {
-  currentLeaderSlotAtom,
-  currentSlotAtom,
-  firstProcessedSlotAtom,
-  leaderSlotsAtom,
-  nextLeaderSlotAtom,
-  slotDurationAtom,
-} from "../../atoms";
+import { useAtomValue } from "jotai";
+import { currentSlotAtom, slotDurationAtom } from "../../atoms";
 import { Box, Flex, Text } from "@radix-ui/themes";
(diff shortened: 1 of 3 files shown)


asuzuki-jumptrading:
@greptile-jt review

src/features/Navigation/SlotsRenderer.tsx
amliu-jumptrading: We can use the atom you've added (`getIsGroupSelectedAtom`) here too

amliu-jumptrading (dismissed):

src/features/Navigation/SlotsRenderer.tsx
```diff
@@ -156,87 +148,84 @@ function CurrentLeaderSlotGroup({ firstSlot }: { firstSlot: number }) {
...
-    </Flex>
-  );
+interface PastSlotGroupProps extends SlotGroupProps {
+  isProcessed: boolean;
+  isSelected: boolean;
 }
-
-function YourProcessedSlotGroup({ firstSlot }: { firstSlot: number }) {
-  const selectedSlot = useAtomValue(selectedSlotAtom);
+function PastSlotGroup({
```
amliu-jumptrading: Can we wrap the function in a `memo` instead of using `useMemo` on the return value?

asuzuki-jumptrading: yes, thanks for catching this! The useMemo was originally to protect against re-renders when accessing `selectedSlot`, but that's not necessary now that we pass in the boolean `isSelected` instead

amliu-jumptrading (approved):

atsai-jumptrading (approved):

PR #367, opened 2026-06-26
fix: stop live sparkline if data doesn't reach right edge

amliu-jumptrading:
The dataWindow and sparkline ticking occur independently of each other. This means if the dataWindow updating falls behind the rightmost-edge of the sparkline ticking (performance.now() when live), a gap will appear between the end of the sparkline and the right edge of the sparkline area. Instead of rendering the next tick, which will show the gap, skip this render.


asuzuki-jumptrading (approved):
Thank you!

PR #369, opened 2026-06-29
chore: fixes for shorter slot times

asuzuki-jumptrading:
- fix shreds deletion range so visible slots don't get prematurely deleted
- turn off integer animations

### DIFF
--- a/src/components/AnimatedInteger.tsx
+++ b/src/components/AnimatedInteger.tsx
@@ -12,6 +12,9 @@ import clsx from "clsx";
 import { Box, Flex, Text, type FlexProps } from "@radix-ui/themes";
 import { useUnmount } from "react-use";
 import useIsDocumentVisible from "../hooks/useIsDocumentVisible";
+import { useAtomValue } from "jotai";
+import { nsPerMs } from "../consts";
+import { epochAtom } from "../atoms";
 
 const MIN_ANIMATION_DURATION_MS = 20;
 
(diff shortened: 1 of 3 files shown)


src/components/AnimatedInteger.tsx
```diff
@@ -29,13 +32,48 @@ interface AnimatedIntegerProps {
...
+  }, [height]);
 
   // re-mount when visible again, so there's no need to catch up
   if (!isDocumentVisible) return null;
 
-  return <AnimatedIntegerInner {...props} />;
+  if (noAnimation) {
+    return (
+      <Flex position="relative" justify={containerRowJustify}>
+        {/* Add hidden Text to set container height and to support selection / copy of the full number (digits components are not selectable). */}
```
amliu-jumptrading: Do we need this comment here? Unlike in AnimatedIntegerInner, this text is visible and not just for text selection.

src/components/AnimatedInteger.tsx
```diff
@@ -29,13 +32,48 @@ interface AnimatedIntegerProps {
...
+        <Text className={clsx(className)} style={heightStyle}>
```
amliu-jumptrading: ✂️  `clsx`

src/components/AnimatedInteger.tsx
asuzuki-jumptrading: oops yes will remove

atsai-jumptrading (approved):

PR #375, opened 2026-06-30
fix: additional memo for slot timeline components

atsai-jumptrading:


asuzuki-jumptrading (approved):
nice!

src/features/Overview/SlotTimeline/index.tsx
```diff
@@ -639,17 +639,24 @@ interface HighlightedSlotBarProps {
...
-}
+    return (
+      <Grid rows="repeat(auto-fill, 1fr)" gap={barTrackGap} flexGrow={flexGrow}>
+        {colors.map((color) => (
+          <MSlotBar key={color} barWidth={barWidth} color={color} />
+        ))}
+      </Grid>
+    );
+  },
+  // colors is a fresh array each render — compare by value
```
amliu-jumptrading, line 654: Why do we have a fresh color array each time? It seems to be because two of the places we're passing in color, we're performing a map. If we trace the array we're mapping `slotBars`, we're getting it from the `useGroupedSlotBars` hook. It doesn't seem like we're using any fields from the returned SlotBarInfo map other than barColor. Can we repurpose that hook to just return barColor instead of SlotBarInfo so we can have a stable colors array and don't need a custom equality function here?

PR #378, opened 2026-07-01
fix: oom from use-debounce timers when app is backgrounded

asuzuki-jumptrading:


src/api/useSetAtomWsData.ts
```diff
@@ -223,6 +226,19 @@ function useUpdateAtoms() {
 
   const [startupTime, setStartupTime] = useAtom(startupTimeAtom);
 
+  const isVisibleRef = useRef(
```
atsai-jumptrading: we have a hook `useIsDocumentVisible`, we should probably do the same and just have a `useIsDocumentVisibleRef` hook for consistency and discoverability

atsai-jumptrading: another option that i feel is cleaner is just to have a global that the `useDebounceIfVIsible` pulls. Could still have an optional ref param but then we don't need to be passing this isVisibleRef everywhere each time we want to use the hook

src/api/useSetAtomWsData.ts
```diff
@@ -873,16 +911,13 @@ function useUpdateAtoms() {
 
   useEffect(() => {
     if (isSocketDisconnected) {
-      dbFlushSupermajorityPeersBuffers.cancel();
```
atsai-jumptrading, line 876: Is there no issue removing this? Why don't we just expose `cancel` from the `useDebounceIfVisible` hook. It can cancel both the debounce/throttle and raf

asuzuki-jumptrading: oh yeah I'll add that back

src/api/useSetAtomWsData.ts
asuzuki-jumptrading: yeah let me add that in a different PR and use it in other places too

asuzuki-jumptrading: Fixed in a different PR and rebased this one

src/api/useDebounceIfVisible.ts
```diff
@@ -0,0 +1,101 @@
...
+
+const store = getDefaultStore();
+
+/**
+ * Only use the throttled callback while visible.
+ * Otherwise, clear the timers and use requestAnimationFrame to call
+ * the callback on the next frame (rAF is paused while backgrounded, so no timers
+ * accumulate).
+ */
+export function useThrottledCallbackIfVisible<
```
amliu-jumptrading, line 15: `useThrottledCallbackIfVisible` and `useDebouncedCallbackIfVisible` are largely implemented the same except for which throttledFn/debouncedFn they use. Would it make sense to implement a `useCallbackIfVisible` with that shared logic and these two functions become wrappers?

amliu-jumptrading (dismissed):

src/api/useDebounceIfVisible.ts
asuzuki-jumptrading: I would prefer to keep these separate for now, since there are only 2 and I don't want to make it more complicated with generic types

atsai-jumptrading (approved):

PR #380, opened 2026-07-01
fix: don't query slots when backgrounded

asuzuki-jumptrading:
When the app is backgrounded but the slots nav is in real-time mode and progressing, we continue to make useSlotQuery requests. Pause those requests until the app is visible again.

### DIFF
--- a/src/hooks/useSlotQuery.ts
+++ b/src/hooks/useSlotQuery.ts
@@ -1,15 +1,16 @@
 import { useAtomValue } from "jotai";
 import {
   getIsFutureSlotAtom,
+  isDocumentVisibleAtom,
   slotPublishAtomFamily,
   slotResponseAtomFamily,
 } from "../atoms";
-import { useMount } from "react-use";
-import { useCallback, useEffect, useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
 import { useWebSocketSend } from "../api/ws/utils";
 import memoize from "micro-memoize";
 import type { SendMessage } from "../api/ws/types";
 import { throttle } from "lodash";
+import { useUnmount } from "react-use";
 
 enum SlotQueryType {
   Publish = "publish",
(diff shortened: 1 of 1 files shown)


src/hooks/useSlotQuery.ts
```diff
@@ -83,15 +91,30 @@ function useSlotQuery(
...
-  useMount(() => {
-    setTimeout(() => setWaitingForData(false), 3_000);
+    queryTimerRef.current = {
+      query,
+      queryTimeoutId: queryTimeout,
+      waitedForDataTimeoutId: waitingTimeout,
+    };
+  }, [query, isDocumentVisible]);
+
+  useUnmount(() => {
```
atsai-jumptrading, line 112: the combination of some things in mount vs useEffect is a bit confusing. I think everything can be moved to the effect and cleanup. Also `queryTimerRef` is somewhat unecessary since the wsSend is gated by a throttle in sendQuery and skipQuery once data is populated. The only thing you gain is robustness over `skipQuery` not being populated correctly

atsai-jumptrading (dismissed):

src/hooks/useSlotQuery.ts
asuzuki-jumptrading: `queryTimerRef` is intended as a way to reset state if `query` changes.
With the current setup, you could change `query` and end up with `waitingForData` not getting reset and the timeouts not getting reset.

I separated the unmount logic from the useEffect because I don't want to cancel the request +  waitingForData timer if a user tabs away and then tabs back into the app with the same query request being needed

amliu-jumptrading (dismissed):

atsai-jumptrading (approved):

src/hooks/useSlotQuery.ts
asuzuki-jumptrading: Discussed offline -- the query timer rescheduling isn't really an issue. We'd still want to reset waitingForData though, so we'll keep the code as is

PR #385, opened 2026-07-07
feat: add accounts page

amliu-jumptrading:
## Cards
- Cache, Disk, Compaction, and Index
- Refactored PieChart used on Overview page to use as a shared component for the new pie charts added to the accounts cards

## Tables
- Cache Classes, Tiles, Partitions
- Genearlized LiveMetricsTable into a DataTable component that owns the table styles and renders column headers from a config


src/components/PieChart.tsx
```diff
@@ -0,0 +1,121 @@
...
+  children,
+  style,
+}: {
+  centerY: number;
+  children: React.ReactNode;
+  style?: React.CSSProperties;
+}) {
+  return (
+    <text
+      y={centerY - 6}
```
asuzuki-jumptrading: How did you decide on 6?

src/components/pieChart.module.css
```diff
@@ -8,16 +8,20 @@ causing a horizontal scrollbar */
 div:has(> .tooltip) {
   width: 1px;
   height: 1px;
+  contain: layout;
```
asuzuki-jumptrading, line 11: What happens visually if we don't have this?

src/utils.ts
```diff
@@ -512,3 +585,30 @@ export function getPeerIconUrl(peerInfo: PeerUpdateInfo | null | undefined) {
...
+  return denominator ? Math.abs(numerator / denominator) * 100 : 0;
+}
+
+// truncate instead of round hit rate values
+export function formatHitRate(fraction: number, precision = 2) {
+  const pct = fraction * 100;
+  return (Math.trunc(pct * 100) / 100).toFixed(precision);
+}
+
+export function formatRate(value: number) {
```
asuzuki-jumptrading: I think we'd want to show 0 values for some rate fields. Could we rename this function to make it more descriptive, or pass in a flag for the 0 being `0` or `-`?

src/features/Accounts/IndexCard/IndexPieChart.tsx
```diff
@@ -0,0 +1,87 @@
...
+  unused: number;
+}) {
+  const data: IndexPieData[] = useMemo(() => {
+    const total = used + unused;
+    return [
+      {
+        id: "used",
+        label: "Used",
+        value: used,
+        color: "#82593F",
```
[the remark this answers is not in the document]
amliu-jumptrading: We're using a slightly different color for the pie segment than the metric. This is also a one-off color so we don't need to export it.

src/features/Accounts/CacheCard/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const copiedPerSec = formatSIBytes(accountStats.io.bytes_copied_per_sec);
+
+  const { numerator: usedBytes, denominator: sizeBytes } =
+    formatIECBytesFraction(
+      cacheClasses.reduce(
+        (a, c) => a + c.used_slots * (cacheClassList[c.class]?.bytes ?? 0),
+        0,
+      ),
+      accountStats.cache.size_bytes,
+      2,
```
asuzuki-jumptrading, line 53: why do we start with 2?

src/features/Accounts/consts.ts
```diff
@@ -0,0 +1,53 @@
+export enum PartitionTier {
+  Hot = 0,
+  Warm = 1,
+  Cold = 2,
+  Off = 255,
```
asuzuki-jumptrading, line 5: Why 255?

src/features/Accounts/DiskCard/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+            <Stat
+              label="Allocated"
+              value={allocated.value}
+              color={accountsSecondaryColor}
+              suffix={allocated.unit}
+              minWidth={storageStatsMinWidth}
+            />
+          </Flex>
+
+          <div style={{ width: 1, background: "#ffffff33" }} />
```
asuzuki-jumptrading: should we extract this color?

src/features/Accounts/DiskCard/DiskPieChart.tsx
```diff
@@ -0,0 +1,108 @@
...
+      value: unusedBytes,
+      color: accountsUnusedColor,
+      pct: getSafePct(unusedBytes, total),
+    },
+  ];
+
+  const tooltipFormatter = (datum: ComputedDatum<DiskPieData>) => {
+    const fmt = formatSIBytes(datum.data.value);
+    return `${fmt.value} ${fmt.unit} (${datum.data.pct.toFixed(1)}%)`;
+  };
```
asuzuki-jumptrading: Should we wrap this in a `useCallback` since we're passing in the needed args?

src/features/Accounts/CompactionCard/index.tsx
```diff
@@ -0,0 +1,185 @@
...
+}) {
+  return (
+    <Flex direction="column" gap="5px">
+      <PartitionUtilization
+        usedFrac={partition.used_frac}
+        fragmentedFrac={partition.fragmented_frac}
+        compactionTriggerFrac={partition.compaction_trigger_frac}
+        isWriteHead={partition.is_write_head}
+        compactionFrac={partition.compaction_frac}
+        compactionState={partition.compaction_state}
```
asuzuki-jumptrading: Why not pass in the whole partition, like with the legend?

src/features/Accounts/CompactionCard/index.tsx
```diff
@@ -0,0 +1,185 @@
...
+              background: "var(--yellow-9)",
+              flexShrink: 0,
+            }}
+          />
+        }
+      />
+      <LegendItem
+        label="Compact Head"
+        value={compactionPct ? fmtPct(compactionPct) : ""}
+        icon={
```
asuzuki-jumptrading: Could we extract and memo / useMemo these icons so they don't need to be re-renderered?

src/features/Accounts/CompactionCard/index.tsx
```diff
@@ -0,0 +1,185 @@
...
+      />
+      <PartitionLegend partition={partition} />
+    </Flex>
+  );
+}
+
+function PartitionLegend({ partition }: { partition: Partition }) {
+  const usedPct = partition.used_frac * 100;
+  const fragPct = partition.fragmented_frac * 100;
+  const triggerPct = partition.compaction_trigger_frac * 100;
```
asuzuki-jumptrading: Would this value change? If not, could we useMemo the formatted value and LegendItem (or `memo` the component)

src/components/TableDescriptionDialog.tsx
```diff
@@ -1,10 +1,16 @@
...
-import { appMaxWidth } from "../../../consts";
+import { appMaxWidth } from "../consts";
+import type { ColumnGroup } from "./DataTable";
 
-export default function TableDescriptionDialog() {
+interface TableDescriptionDialogProps {
+  groups: ColumnGroup[];
+}
+
+export default function TableDescriptionDialog({
```
asuzuki-jumptrading: Let's memo this component so it doesn't re-render when parent data atoms update but groups stay stable

src/features/Accounts/CacheClasses/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+          <CacheSpace
+            usedBytes={usedBytes}
+            usedPct={usedPct}
+            unusedPct={unusedPct}
+            color={color}
+            unusedColor={unusedColor}
+          />
+        ) : (
+          ""
+        )}
```
asuzuki-jumptrading: nit
```suggestion
        {!!overallMaxBytes && (
          <CacheSpace
            usedBytes={usedBytes}
            usedPct={usedPct}
            unusedPct={unusedPct}
            color={color}
            unusedColor={unusedColor}
          />
        ) }
```

src/features/Accounts/Partitions/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+        <Flex align="center" gap="5px">
+          <CardHeader text="Partitions" />
+          <TableDescriptionDialog groups={partitionGroups} />
+        </Flex>
+        <DataTable
+          groups={partitionGroups}
+          TableBody={PartitionsTableBody}
+          data={accountStats}
+          hideGroupHeaders
+        />
```
asuzuki-jumptrading: We could reduce re-renders of `DataTable` siblings by using a different pattern:

- create an atom `hasAccountStats` that returns `!!accountStats`. We could even move this to Accounts/index.tsx instead of each card.
- Remove `data` prop from DataTable. Instead, have the `TableBody` like `PartitionsTableBody` itself read from atoms to get the data.

src/features/Gossip/StorageStatsCharts.tsx
```diff
@@ -130,81 +140,47 @@ function StorageStatsPieChart({
         .map((node, i) => {
           return { ...node, color: getColor(i) };
         }),
-    ];
+    ] as StoragePieData[];
```
asuzuki-jumptrading: nit:
```suggestion
    ] satisfies StoragePieData[];
```

src/features/Accounts/Tiles/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+function DataRow({ tile }: { tile: Tile }) {
+  const hasReads =
+    tile.acquired_per_sec > 0 ||
+    tile.acquired_writable_per_sec > 0 ||
+    tile.read_ops_per_sec > 0;
+
+  return (
+    <Table.Row className={tableStyles.dataRow}>
+      <Table.Cell>{tile.joiner_type}</Table.Cell>
```
asuzuki-jumptrading: We could prevent updates to some of these unchanged cells by creating a component that wraps Table.Cell with `memo`

src/features/Accounts/Tiles/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+          <DataRow key={`${tile.name}:${tile.kind_id}`} tile={tile} />
+        ),
+      )}
+    </Table.Body>
+  );
+}
+
+type Tile = AccountsStats["tiles"][number];
+
+function PinnedRow({ tile }: { tile: Tile }) {
```
asuzuki-jumptrading: Could we `memo` this to prevent re-renders when other fields change but the name + kind don't?

src/components/PieChart.tsx
amliu-jumptrading: It came from the CenteredMetric in the ValidatorStatsChart I generalized this component from, but we should actually just let the children determine their y position. I'm taking this out and just passing `y={centerY}` and adjusting the `dy` values in the `tspans` of the children.

src/components/pieChart.module.css
amliu-jumptrading: Without this, if the tooltip extends off the page, it will affect the layout and add a scrollbar to the page (reproduce by hover over the validator pie chart on the gossip page). With this, the tooltip's internal layout is isolated from the rest of the page and would affect the page layout.

src/features/Accounts/consts.ts
amliu-jumptrading: These are the values that the websocket sends

> Compaction tier: 0 (hot), 1 (warm), 2 (cold), or 255 (off, fully compacted and awaiting reclaim)

src/features/Accounts/CacheCard/index.tsx
amliu-jumptrading: For the reduce we start with 0. The 2 is a precision prop for the `formatIECBytesFraction` function

src/components/pieChart.module.css
asuzuki-jumptrading: I see, thanks for explaining!

src/features/Accounts/CacheCard/index.tsx
asuzuki-jumptrading: ahh ok sorry I missed that. Thanks!

src/features/Accounts/consts.ts
asuzuki-jumptrading: Ah ok. Could we move this to the entities file, by the `accountsPartitionSchema`? We could also create a z.enum schema for it

src/features/Accounts/CompactionCard/index.tsx
```diff
@@ -0,0 +1,185 @@
...
+    </Flex>
+  );
+}
+
+function LegendItem({
+  icon,
+  label,
+  value,
+}: {
+  icon: ReactNode;
```
atsai-jumptrading: why not just make this `children`? Makes the consumer not have to pass a component into a prop

atsai-jumptrading (dismissed):

asuzuki-jumptrading (dismissed):

src/utils.ts
amliu-jumptrading: I'll add a flag `showZero` that defaults to `false`

src/features/Accounts/DiskCard/index.tsx
amliu-jumptrading: This is a one-off color we're not using elsewhere that also doesn't have any semantic meaning, so I don't want to add it to to colors.ts. I can move it into a const at the top of the file though

src/features/Accounts/DiskCard/DiskPieChart.tsx
amliu-jumptrading: Looks like it doesn't actually use anything local within the component, so I'll just move it out to the module level

src/features/Accounts/CompactionCard/index.tsx
amliu-jumptrading: I've modified `LegendItem` to take in `icon` as `children` and since they are static divs, React's reconciler will skip DOM updates for them

src/features/Accounts/CompactionCard/index.tsx
amliu-jumptrading: Fair, I'll just pass in the whole partition

src/features/Accounts/CompactionCard/index.tsx
amliu-jumptrading: I don't think it's worth it to memo these values and `LegendItem`. Memoing `LegendItem` won't help because when the `PartitionLegend` rerenders, the icons will have a new object reference each render and fail the shallow prop comparison. If we memoed the icons so they were a stable reference, it would work, but I think the additional overhead of the dependency array and cache lookup for the memoization would be equivalent to cheap operations like rounding, string formatting, and rendering a static div.

src/features/Accounts/Partitions/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+        {formatRate(partition.read_ops_per_sec)}
+      </Table.Cell>
+      <Table.Cell align="right" className={accountStyles.write}>
+        {formatRate(partition.write_ops_per_sec)}
+      </Table.Cell>
+      <Table.Cell align="right" className={accountStyles.read}>
+        {formatSIBytesRate(partition.bytes_read_per_sec)}
+      </Table.Cell>
+      <Table.Cell align="right" className={accountStyles.write}>
+        {formatSIBytesRate(partition.bytes_written_per_sec)}
```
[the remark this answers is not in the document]
amliu-jumptrading: Modified the descriptions since the data itself is correct

src/components/DataTable.tsx
```diff
@@ -0,0 +1,184 @@
...
+}: InnerTableProps) {
+  const groups = useMemo(
+    () => allGroups.filter(({ pinned }) => !!pinned === !!isPinned),
+    [allGroups, isPinned],
+  );
+
+  const rootStyle = useMemo(() => {
+    if (!isPinned) return { ...style, minWidth: "0px" };
+
+    // start with one pixel to account for border width
```
atsai-jumptrading, line 78: can we change the box sizing or something instead?

src/features/Accounts/Stat.tsx
```diff
@@ -0,0 +1,77 @@
...
+type Size = "sm" | "lg";
+
+interface StatProps {
+  className?: string;
+  label?: string;
+  value: string;
+  size?: Size;
+  color?: CSSProperties["color"];
+  suffix?: string;
+  suffixColor?: CSSProperties["color"];
```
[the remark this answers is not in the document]
amliu-jumptrading: It was used before, but no longer applied with a design update, so I can remove

src/components/PieChart.tsx
```diff
@@ -0,0 +1,121 @@
...
+  enableArcLabels = false,
+  arcLabelsSkipAngle,
+  enableArcLinkLabels = false,
+  arcLinkLabelsSkipAngle,
+  arcLabelsTextColor,
+  arcLabel,
+}: PieChartProps<D>) {
+  const layers: PieLayer<D>[] = ["arcs"];
+  if (enableArcLabels) layers.push("arcLabels");
+  if (centeredMetric) layers.push(centeredMetric);
```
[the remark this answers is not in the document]
amliu-jumptrading: arc link labels were not actually being used in any of the charts, so it's safe to remove that prop.

src/features/Accounts/Partitions/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+      </Table.Cell>
+      <Table.Cell align="right">
+        {isCompacting ? `${compactionPct.toFixed(1)}%` : "-"}
+      </Table.Cell>
+      <Table.Cell align="right">
+        {formatDuration(partition.age_seconds)}
+      </Table.Cell>
+      <Table.Cell align="right">
+        {formatDuration(partition.filled_seconds)}
+      </Table.Cell>
```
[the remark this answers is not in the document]
amliu-jumptrading: This is fine as is

atsai-jumptrading (approved):

src/features/Accounts/CompactionCard/index.tsx
asuzuki-jumptrading: Yeah the formatting is cheap, but if `partition` is constantly changing and the formatted `value` prop that we pass into `LegendItem` isn't (maybe due to rounding) it could still be worth it to memo LegendItem. We could move the icons outside of the component to prevent new object references every time partition updates.

amliu-jumptrading: I still think the benefit is marginal since `LegendItem` is only 2 text components and the icon. Memoizing `LegendItem` comes with the cost of the check for label, value, and the icon/children that is about the same as re-rendering it. It could be more worth it if `LegendItem` is a heavier component to re-render, but right now it's preemptive.

What's an example of when rounding/the formatted value prop would be expensive. Usually I consider rounding and string interpolation to be pretty cheap.

src/components/DataTable.tsx
amliu-jumptrading: Box sizing doesn't work here because that relates to the element's own border. The right border that is getting cropped without this is the border of the column group itself which is inside of the table that this style gets applied to.

src/features/Accounts/CompactionCard/index.tsx
asuzuki-jumptrading: The rounding / formatting is not expensive. And yes, the benefit could be negligible depending on the data behavior but I think it's worth changing for the current data I see.

I see that `partition` is updated every 100ms. But the `LegendItem` value rarely changes in Idle state. Frag changes once every ~30 seconds, and I'm guessing the other LegendItems change every half hour or so. It could be because the numerical pcts change so slightly / not often, and so the rounded pcts barely change.

If we memo `LegendItem` and pull out the icons that don't depend on props, we can remove the unnecessary re-renders we're doing every 100ms and just do them every 30 seconds or less. Even though the component themselves are light, I think it's better to eliminate those extra renders because they happen at a sub-second frequency.

asuzuki-jumptrading: I think the better approach here would be to memo The parent `PartitionLegend`, and then only update all the legend items once every 30 seconds. That would be better than doing the memo comparisons for every LegendItem, which is a small win compared to their actual render logic

PR #389, opened 2026-07-09
fix: optional tile metrics

asuzuki-jumptrading:


atsai-jumptrading (approved):

asuzuki-jumptrading:
We'll be adding these fields to Frankendancer.
Also, we'd want to hide the columns if they are not available.

PR #393, opened 2026-07-16
chore: sync live metric tiles pinned row with data row updates

amliu-jumptrading:
The pinned row and data row in the live metric tiles table were updating out of sync because the pinned row was on the react render cycle while the data row was subscribing to atom updates and changing the text content of the elements directly in a useLayoutEffect. This causing the table to "flash" when a tile shutdown happens and a new one up to replace it (observed frequently with netlnk) because either the pinned row or the data row would render before the other leaving an incomplete row for a single frame. This syncs their updates by using the same method to subscribe to and update the data.


asuzuki-jumptrading (approved):
Nice!

PR #397, opened 2026-07-24
feat: add version to leader schedule search

devinschmitz:
Adds gossip version as a search filter on the leader schedule page.

image

image


asuzuki-jumptrading (approved):
Thank you!

PR #398, opened 2026-07-27
chore: migrate latency_exact/is_voter fields

jherrera-jump:


src/atoms.ts
amliu-jumptrading: Instead of having Frankendancer/Firedancer branches of logic within these atoms and taking `latency_exact` as a third argument when updating the`_lateVotesMapAtom` (for `addLateVoteSlotAtom` and `setLateVoteHistoryAtom`), can we separate them entirely into factory functions?

Both factories will return the same atoms, but the logic inside them will differ
```
export const [
  discountedLateVoteSlotsAtom,
  addLateVoteSlotAtom,
  deleteLateVoteSlotAtom,
  clearLateVoteSlotsAtom,
  setLateVoteHistoryAtom,
] = isFrankendancer
  ? getFrankendancerLateVoteSlotsAtom()
  : getFiredancerLateVoteSlotsAtom();
```

We can create a factory `getFrankendancerLateVoteSlotsAtom` that returns the atoms from before your changes. This keeps the Frankendancer logic separated from the Firedancer logic and makes it easy to remove once we stop Frankendancer support.

The main benefit of using a factory though comes from adding `getFiredancerLateVoteSlotsAtom`. Since `latency_exact` is already discounted, we don't need to store the vote latencies as a map (this was necessary when we were calculating the discounted latency client-side). Instead, we can store it as a set and directly return that for `discountedLateVoteSlotsAtom` without unnecessarily subscribing to `skippedClusterSlotsAtom`.
```
const _lateVoteSlotsAtom = atomWithImmer(new Set<number>());
```

amliu-jumptrading: So that the interface for `setLateVoteHistory` is the same for both factories, you'll also want to branch what you pass through as `latency` in `useSetAtomWsData`

```
setLateVoteHistory(
    isFiredancer
        ? { slot: value.slot, latency: value.latency_exact ?? [] }
        : { slot: value.slot, latency: value.latency ?? [] }
);
```

src/utils.ts
```diff
@@ -557,10 +557,17 @@ export function formatTimeNanos(
...
+    if (publish.skipped) return false;
+    return (
+      publish.vote_latency_exact === null || publish.vote_latency_exact > 1
+    );
+  }
   return (
-    publish.level === "rooted" &&
-    ((publish.vote_latency && publish.vote_latency > 1) ||
-      (publish.vote_latency === null && !publish.skipped))
+    (publish.vote_latency !== null && (publish.vote_latency ?? 1) > 1) ||
```
amliu-jumptrading: Since we check for `publish.vote_latency !== null`, we don't need the null coalescing operator here. We can just use `(publish.vote_latency !== null && publish.vote_latency > 1)`

src/features/LeaderSchedule/Slots/SlotCardGrid.tsx
```diff
@@ -248,20 +248,27 @@ function getRowValues(
...
+          )
+        : null;
+
+  const voteLatency =
+    publish.is_voter === false
+      ? { text: "-" }
+      : discountedLatency != null
+        ? { text: discountedLatency.toLocaleString() }
+        : publish.skipped
+          ? { text: "-" }
```
amliu-jumptrading, line 268: Is it intentional that skipped shows "-" now? Before we showed "".

jherrera-jump: yeah I think it makes more sense to treat "structurally cannot vote for this slot" (i.e. skipped or is_voter===false) with "-". And treat "not known yet" as "". Forgot to change the latter will fix.

jherrera-jump: Actually "" doesn't look great for the latter. So lets just use "-" for both. Not that important to distinguish the two edge cases.

src/atoms.ts
```diff
@@ -863,33 +863,34 @@ export const serverTimeMsAtom = atom((get) => {
...
+type FiredancerLateVoteHistory = {
+  slot: number[];
+  latency_exact: (number | null)[];
+};
+
+type LateVoteHistory = FrankendancerLateVoteHistory | FiredancerLateVoteHistory;
+
+type LateVoteSlotsRange = { startSlot: number; endSlot: number };
+
+export function getFrankendancerLateVoteSlotsAtom() {
```
amliu-jumptrading: We can remove the export here since we're using it internally on this page.

src/atoms.ts
```diff
@@ -926,22 +927,85 @@ export const [
...
+          newLateVotesMap.set(slot, value.latency[expandedSlotIndex] ?? null);
+          expandedSlotIndex++;
+        }
+      }
+      set(_lateVotesMapAtom, newLateVotesMap);
+    }),
+  ] as const;
+}
+
+export function getFiredancerLateVoteSlotsAtom() {
```
amliu-jumptrading: Same here remove the export

src/atoms.ts
```diff
@@ -926,22 +927,85 @@ export const [
...
+        const runStart = value.slot[run * 2];
+        const runEnd = value.slot[run * 2 + 1];
+        for (let slot = runStart; slot <= runEnd; slot++) {
+          newLateVotesMap.set(slot, value.latency[expandedSlotIndex] ?? null);
+          expandedSlotIndex++;
+        }
+      }
+      set(_lateVotesMapAtom, newLateVotesMap);
+    }),
+  ] as const;
```
amliu-jumptrading: These changes can get reverted since the logic for the Frankendancer case hasn't changed. Personally I find the latency-encoding more intuitive as incrementing through the value.slots array in pairs vs run length encoding (which makes perfect sense for the Firedancer case).

src/atoms.ts
```diff
@@ -926,22 +927,85 @@ export const [
...
-      null,
-      (_get, set, value: { slot: number[]; latency: (number | null)[] }) => {
-        const newLateVotesMap = new Map<number, number | null>();
-        let latencyIdx = 0;
-        for (let i = 0; i < value.slot.length; i += 2) {
-          for (let slot = value.slot[i]; slot <= value.slot[i + 1]; slot++) {
-            newLateVotesMap.set(slot, value.latency[latencyIdx]);
-            latencyIdx++;
+    atom(null, (_get, set, value: LateVoteHistory) => {
+      if (!("latency" in value)) return;
```
amliu-jumptrading: It would be cleaner move this check to the call site and keep the interface for this atom as `{ slot: number[]; latency: (number | null)[] }`

Instead in `useSetAtomWsData`, you can
```
if ((isFrankendancer && vote.latency) || (!isFrankedancer && vote.latency_exact)) {
    setLateVoteHistory(value)
}
```

src/atoms.ts
```diff
@@ -926,22 +927,85 @@ export const [
...
-        set(_lateVotesMapAtom, newLateVotesMap);
-      },
-    ),
-  ];
-})();
+      });
+    }),
+
+    atom(null, (_get, set, value: LateVoteHistory) => {
+      if (!("latency_exact" in value)) return;
```
amliu-jumptrading: Same as above. It would be cleaner to do this check at the callsite.

src/atoms.ts
```diff
@@ -940,8 +927,80 @@ export const [
...
+  discountedLateVoteSlotsAtom,
+  addLateVoteSlotAtom,
+  deleteLateVoteSlotAtom,
+  clearLateVoteSlotsAtom,
+  setLateVoteHistoryAtom,
+] = isFrankendancer ? frankendancerLateVoteAtoms : firedancerLateVoteAtoms;
+
+export const setFrankendancerLateVoteHistoryAtom =
+  frankendancerLateVoteAtoms[4];
+export const setFiredancerLateVoteHistoryAtom = firedancerLateVoteAtoms[4];
```
amliu-jumptrading: ✂️ These don't seem to be used anywhere. Let's remove them.

amliu-jumptrading (dismissed):

PR #399, opened 2026-07-28
feat: estimate next compaction time

amliu-jumptrading:
- calculate next compaction time
  - prioritize in order of state idle < queued < compacting
  - within the same state, prioritize by smaller estimated time
- only show 0s if duration is actually 0
- add defaults for when next compaction partition is undefined so layout does not shift when it is missing

### DIFF
--- a/src/atoms.ts
+++ b/src/atoms.ts
@@ -8,6 +8,7 @@ import {
   serverTimeNanosAtom,
   skippedSlotsAtom,
   startupProgressAtom,
+  accountsStatsAtom,
 } from "./api/atoms";
 import type {
   Epoch,
@@ -17,10 +18,12 @@ import type {
   SlotLevel,
   SlotResponse,
   SupermajorityEpoch,
+  AccountsPartition,
 } from "./api/types";
 import { clamp, merge } from "lodash";
 import {
   getDiscountedVoteLatency,
+  getDurationText,
   getLeaderSlots,
   getSlotGroupLeader,
   getStake,
@@ -30,7 +33,9 @@ import { selectedSlotAtom } from "./features/Overview/SlotPerformance/atoms";
 import { atomFamily } from "jotai/utils";
 import memoize from "micro-memoize";
 import { isFrankendancer } from "./client";
+import { CompactionState } from "./api/entities";
 import { numQuickSearchSlots } from "./features/SlotDetails/const";
+import { Duration } from "luxon";
 
 export const isDocumentVisibleAtom = atom<boolean>(
   document.visibilityState === "visible",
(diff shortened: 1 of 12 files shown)


src/atoms.ts
```diff
@@ -1018,3 +1022,68 @@ export const quickSearchSlotsAtom = atom((get) => {
...
+        // Excludes age_seconds to avoid noisy updates for newly active write heads
+        const hasChanged =
+          prev.size !== new Set(writeHeads.map((p) => p.tier)).size ||
+          writeHeads.some((p) => {
+            const snap = prev.get(p.tier)?.partition;
+            return (
+              snap?.used_frac !== p.used_frac ||
+              snap.fragmented_frac !== p.fragmented_frac ||
+              snap.compaction_state !== p.compaction_state ||
+              snap.compaction_trigger_frac !== p.compaction_trigger_frac
```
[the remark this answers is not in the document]
amliu-jumptrading: I've refactored it so that instead of using the partition from this atom, we return the partition idx and use the partition from the accountsStats atom (which will have the most up to date fields always)

src/features/Accounts/CompactionCard/useNextCompaction.ts
```diff
@@ -0,0 +1,46 @@
...
+import { getDurationText } from "../../../utils";
+import { Duration } from "luxon";
+import { CompactionState } from "../../../api/entities";
+
+export interface NextCompaction {
+  partitionIdx: number;
+  timeLabel: string;
+}
+
+export function useNextCompaction(): NextCompaction | undefined {
```
asuzuki-jumptrading: Would we use this in more than once place on the page? If so, we can move it into a derived atom instead of a hook.

src/features/Accounts/CompactionCard/index.tsx
```diff
@@ -8,58 +8,90 @@ import { useAtomValue } from "jotai";
...
+import { partitionTierLabel, partitionTierColor } from "../consts";
+import { CompactionState } from "../../../api/entities";
 
 function fmtPct(pct: number) {
   return `${Math.round(pct)}%`;
 }
 
 export default function CompactionCard({ className }: { className?: string }) {
   const accountStats = useAtomValue(accountsStatsAtom);
+  const nextCompaction = useNextCompaction();
```
asuzuki-jumptrading: This will run with any accountsStats change, even if `accountsNextCompactionCandidates` doesn't change. So we may want to move it into a derived atom or useMemo some values inside `useNextCompaction`

asuzuki-jumptrading (dismissed):

src/features/Accounts/CompactionCard/useNextCompaction.ts
amliu-jumptrading: Not on the same page, but we'll also use it for the overview accounts card once we add it. We're not using the `accountsNextCompactionCandidates` values directly anywhere except this hook, so I'll simply refactor that atom to return the next compaction values instead of the candidates.

asuzuki-jumptrading (approved):

PR #401, opened 2026-07-30
feat: accounts card

amliu-jumptrading:
- add accounts card to overview page
- update overview card layout (separate frankendancer vs firedancer layout entirely with different breakpoints)

### DIFF
--- a/src/features/Overview/AccountsCard/index.tsx
+++ b/src/features/Overview/AccountsCard/index.tsx
@@ -1,0 +1,166 @@
+import { Flex } from "@radix-ui/themes";
+import Card from "../../../components/Card";
+import CardHeader from "../../../components/CardHeader";
+import { accountsStatsAtom } from "../../../api/atoms";
+import { useAtomValue } from "jotai";
+import Stat from "../../Accounts/Stat";
+import { formatCount, formatHitRate, formatSIBytes } from "../../../utils";
+import {
+  accountsReadColor,
+  accountsWriteColor,
+  accountsUsedColor,
+  accountsFragmentedColor,
+  accountsSecondaryColor,
+} from "../../../colors";
+import styles from "./accountsCard.module.css";
+import { hitRateChangedColor, getHitRateStatus } from "../../../hitRate";
+import { accountsNextCompactionAtom } from "../../../atoms";
+
+const storageStatsMinWidth = "150px";
+const readWriteStatsMinWidth = "120px";
+const cacheStatsMinWidth = "80px";
+
+export default function AccountsCard({ className }: { className?: string }) {
+  const accountStats = useAtomValue(accountsStatsAtom);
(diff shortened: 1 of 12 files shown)


src/utils.ts
```diff
@@ -630,3 +630,17 @@ export function formatSIBytesRate(value: number) {
...
+
+export function getColorByThreshold(
+  thresholds: ColorThreshold[],
+  fallback: string,
+): (value: number) => string {
+  return (value) => {
+    for (const { below, color } of thresholds) {
+      if (value < below) return color;
+    }
+    return fallback;
```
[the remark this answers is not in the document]
amliu-jumptrading: This is a relatively simple function. The tests we currently cover handle functions with more complex logic.

src/features/Accounts/consts.ts
```diff
@@ -46,3 +52,11 @@ export const cacheClassUnusedColors = [
...
   "var(--red-3)",
 ];
+
+export const getCacheHitRateColor = getColorByThreshold(
+  [
+    { below: 0.99, color: badChangedColor },
+    { below: 0.995, color: averageChangedColor },
+  ],
+  goodChangedColor,
+);
```
asuzuki-jumptrading: I think it would be simpler to just have a function with if / else if / else checks unless we're planning to have many color thresholds across the app

src/features/Overview/AccountsCard/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+  return (
+    <Card className={className}>
+      <Flex direction="column" gap="2">
+        <CardHeader text="Accounts" />
+
+        <Flex gap="3" wrap="wrap">
+          <Flex direction="column" gap="2" flexGrow="1">
+            <div className={styles.sectionLabel}>Cache</div>
+            <Flex direction="column" gap="2">
```
asuzuki-jumptrading: Do we need this second flex column inside a flex column?

asuzuki-jumptrading (dismissed):

atsai-jumptrading (dismissed):

src/features/Accounts/consts.ts
amliu-jumptrading: The program cache has a different color threshold than the accounts. Let me check whether we want to consolidate those

src/features/Overview/AccountsCard/index.tsx
amliu-jumptrading: Good catch, it's not needed

src/features/Accounts/consts.ts
amliu-jumptrading: I consolidated the thresholds for the program cache card, accounts card, and hit rates across the accounts page. I've also simplified it to just use if statements
Bad: < 0.99
Average: < 0.995
Good

asuzuki-jumptrading (approved):

PR #404, opened 2026-08-04
refactor: canvas tps chart

amliu-jumptrading:
- rewrites TPS chart from SVG to canvas
- draws on every animation frame instead of every 100ms tick when tps data is pushed
- fix syncing of tps data pushing (100ms) with the estimatedTpsAtom updating (previously 400ms)
- fix the assumed sampling rate of tpsHistory points (previously 400ms)
- derive window label from window ms

Related backend change to account for tps history samples being 200ms apart: https://github.com/firedancer-io/firedancer/commit/73146bf3e5831d902cef59ca8d941294cd66d310

### DIFF
--- a/src/features/Overview/TransactionsCard/Chart.tsx
+++ b/src/features/Overview/TransactionsCard/Chart.tsx
@@ -1,7 +1,6 @@
-import AutoSizer from "react-virtualized-auto-sizer";
-import { useMemo, useRef } from "react";
-import { isDefined } from "../../../utils";
-import { useAtomValue } from "jotai";
+import { useLayoutEffect, useRef } from "react";
+import { getDefaultStore } from "jotai";
+import { useMeasure, useRafLoop } from "react-use";
 import { tpsDataAtom } from "./atoms";
 import {
   regularTextColor,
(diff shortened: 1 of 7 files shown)


src/features/Overview/TransactionsCard/index.tsx
```diff
@@ -17,11 +20,17 @@ export default function TransactionsCard({
...
+              minWidth="180px"
+              minHeight="80px"
+              overflow="hidden"
+              position="relative"
+            >
               <Chart />
             </Box>
             <Flex justify="between">
-              <Text className={styles.axisText}>~ 1min ago</Text>
+              <Text className={styles.axisText}>~ {WINDOW_LABEL}</Text>
```
asuzuki-jumptrading: Will this window label be changing a lot? If it's usually around 1 min, I think we can keep it as `~ 1 min ago`

src/features/Overview/TransactionsCard/Chart.tsx
```diff
@@ -1,131 +1,135 @@
...
+    drawArea(ctx, points, width, height, "voteY", transactionVotePathColor);
+
+    drawPeakLine(ctx, width, height, maxTotalY * yRatio, maxTotalY);
+  });
 
   return (
-    "M" +
-    path.slice(1) +
-    `L ${points[points.length - 1].x} ${height} L ${points[0].x} ${height}, L ${points[0].x} ${points[0].y}`
+    <div ref={measureRef} style={{ position: "absolute", inset: 0 }}>
```
asuzuki-jumptrading, line 81: Why does this need to be position: absolute?

asuzuki-jumptrading (dismissed):
Nice!

src/features/Overview/TransactionsCard/index.tsx
amliu-jumptrading: Fair enough, it was helpful during dev when I was adjusting the window_ms but it's static in practice. I'll just add a comment by WINDOW_MS that this label needs to be updated too if the value changes

src/features/Overview/TransactionsCard/Chart.tsx
amliu-jumptrading: We make it `position: absolute` and `inset: 0` so that its height is determined by the parent `Box` instead of its content `canvas`. Then we measure the div's dimensions to pass into the canvas.

src/features/Overview/TransactionsCard/consts.ts
```diff
@@ -1 +1,5 @@
-export const maxTransactionChartPoints = 600;
+// Frequency server samples history
+export const HISTORY_INTERVAL_MS = 200;
+
+// Keep in sync with the hardcoded "~ 1min ago" axis label in index.tsx
+export const WINDOW_MS = 60_000;
```
asuzuki-jumptrading, line 2: We could update the ~ 1min ago text to use this `WINDOW_MS` value so we're only hard-coding this window

asuzuki-jumptrading (dismissed):

src/features/Overview/TransactionsCard/consts.ts
asuzuki-jumptrading: Sorry, I misunderstood what your code was doing previously!
I thought it was using the first ts in the data and showing that in the label, when in fact it was using WINDOW_MS.
You can ignore the above comment 😅 If we add a feature in the future to be able to control the window size, we can pull in the change you had previously.

atsai-jumptrading (dismissed):

amliu-jumptrading:
I've consolidated the `tpsSampleIntervalMs` and `HISTORY_INTERVAL_MS` because the live updates come at the same rate that the history sample points are spaced. `tpsSampleIntervalMs` should be `200ms`

atsai-jumptrading (approved):

PR #410, opened 2026-08-19
feat: replay page

asuzuki-jumptrading:
- set up replay page, which checks for WebGL support
- hide left nav

### DIFF
--- a/src/hooks/useSlotsNavigation.ts
+++ b/src/hooks/useSlotsNavigation.ts
@@ -7,17 +7,22 @@ import { useMedia } from "react-use";
 export function useSlotsNavigation() {
   const isNarrowScreen = useMedia(narrowNavMedia);
   const [isNavCollapsed, setIsNavCollapsed] = useAtom(_isNavCollapsedAtom);
-  const isLeaderSchedule = useCurrentRoute() === "Schedule";
+  const currentRoute = useCurrentRoute();
 
-  const showNav = isLeaderSchedule || !isNavCollapsed;
-  const showOnlyEpochBar = isLeaderSchedule;
+  const noNav = currentRoute === "Replay";
+  const showOnlyEpochBar = currentRoute === "Schedule";
+
+  const showNav = !noNav && (showOnlyEpochBar || !isNavCollapsed);
 
   return {
+    noNav,
     isNarrowScreen,
     showNav,
     setIsNavCollapsed,
     showOnlyEpochBar,
-    blurBackground: isNarrowScreen && !isNavCollapsed && !showOnlyEpochBar,
-    occupyRowWidth: showOnlyEpochBar || (!isNarrowScreen && !isNavCollapsed),
+    blurBackground:
+      !noNav && isNarrowScreen && !isNavCollapsed && !showOnlyEpochBar,
+    occupyRowWidth:
+      !noNav && (showOnlyEpochBar || (!isNarrowScreen && !isNavCollapsed)),
   };
 }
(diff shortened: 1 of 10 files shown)


src/hooks/useSlotsNavigation.ts
```diff
@@ -9,15 +9,19 @@ export function useSlotsNavigation() {
...
   return {
+    noNav,
     isNarrowScreen,
     showNav,
     setIsNavCollapsed,
     showOnlyEpochBar,
     blurBackground: isNarrowScreen && !isNavCollapsed && !showOnlyEpochBar,
-    occupyRowWidth: showOnlyEpochBar || (!isNarrowScreen && !isNavCollapsed),
+    occupyRowWidth:
+      !noNav && (showOnlyEpochBar || (!isNarrowScreen && !isNavCollapsed)),
```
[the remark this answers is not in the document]
asuzuki-jumptrading: Technically you can't expand the slots nav in mobile view on the replay page, but I will add this check to the `blurBackground` field

asuzuki-jumptrading:
> ## Pull request overview
> Copilot reviewed 8 out of 8 changed files in this pull request and generated no new comments.
> 
> Suppressed comments (1)
> **src/features/Replay/index.tsx:7**
> 
> * `isWebgl2SupportedAtom` is not a pure capability check: it is also set to `false` after renderer setup or context-restoration failures and is never reset. A transient failure in the Overview chart therefore makes Replay unavailable for the rest of the session, even after that chart is unmounted and the GPU recovers. Keep immutable WebGL2 capability separate from per-renderer failure/fallback state, and let Replay handle failures from its own renderer.
> 
> ```
>   const isWebGl2Supported = useAtomValue(isWebgl2SupportedAtom);
> ```

This is ok for now... the user could refresh the page, and we can create a separate atom for replay webgl2 support if needed

src/hooks/useSlotsNavigation.ts
```diff
@@ -9,15 +9,20 @@ export function useSlotsNavigation() {
   const [isNavCollapsed, setIsNavCollapsed] = useAtom(_isNavCollapsedAtom);
   const isLeaderSchedule = useCurrentRoute() === "Schedule";
 
-  const showNav = isLeaderSchedule || !isNavCollapsed;
+  const noNav = useCurrentRoute() === "Replay";
```
amliu-jumptrading: Should we have a `const router= useCurrentRoute()` call once and use that for both `isLeaderSchedule` and `noNav` checks?

src/hooks/useSlotsNavigation.ts
```diff
@@ -9,15 +9,20 @@ export function useSlotsNavigation() {
...
+
+  const showNav = !noNav && (isLeaderSchedule || !isNavCollapsed);
   const showOnlyEpochBar = isLeaderSchedule;
```
amliu-jumptrading: nit: Can we move the `showOnlyEpochBar` const above `showNav` and use it instead of `isLeaderSchedule` so it reads more contextually as "show nav when there is not no nav AND only the epoch bar shows or the nav is not collapsed". This also makes `showOnlyEpochBar` the one source of truth rather than needing to update both if there is another page that should only show the epoch bar.

I realize this is existing 😅 and a small detail , but wanted to point it out since the `showNav` was changing

amliu-jumptrading (approved):

atsai-jumptrading (approved):

amliu-jumptrading (approved):

PR #411, opened 2026-08-20
feat: aggregate fees track

asuzuki-jumptrading:
- set up replay Chart, which keeps track of world and visible range of all tracks
- set up fees track (currently only aggregate data)


src/features/Replay/RevenueTrack/utils.ts
```diff
@@ -0,0 +1,212 @@
...
+import type { RevenueType } from "../../../api/entities.ts";
+import type { AggRevenue } from "../../../api/types.ts";
+import { omit } from "lodash";
+import { getRevenueRatio } from "../../../mathUtils.ts";
+
+// TODO: set reasonable threshold with non-agg data
+const AGGREGATE_THRESHOLD_MS = 0;
+
+const REVENUE_COLOR: RgbColor = [116 / 255, 178 / 255, 238 / 255];
+const REVENUE_OPACITY = 0.8;
```
amliu-jumptrading: Why do we have opacity for aggregated revenue? Won't there be no overlap and thus no need for opacity?

src/api/atoms.ts
```diff
@@ -200,3 +203,7 @@ export const liveProgramCacheAtom = atom<LiveProgramCache | undefined>(
 export const healthAtom = atom<Health | undefined>(undefined);
 
 export const accountsStatsAtom = atom<AccountsStats | undefined>(undefined);
+
+export const aggSlotsAtom = atom<AggSlots | undefined>(undefined);
+export const aggRevenueAtom = atom<AggRevenue | undefined>(undefined);
+export const aggShredsAtom = atom<AggShreds | undefined>(undefined);
```
amliu-jumptrading: nit: can we remove slots and shreds since this pr is scoped to only add the fees track

src/api/entities.ts
```diff
@@ -1226,3 +1230,78 @@ export const accountsSchema = z.discriminatedUnion("key", [
...
+  }),
+  timelineTopicSchema.extend({
+    key: z.literal("query_agg_shreds"),
+    value: aggShredsSchema,
+  }),
+  timelineTopicSchema.extend({
+    key: z.literal("query_shreds"),
+    value: liveShredsSchema,
+  }),
+]);
```
amliu-jumptrading: nit: same here, can we remove slots and shreds since they're not being used for this feature

src/api/types.ts
```diff
@@ -216,3 +220,8 @@ export type AccountsStats = z.infer<typeof accountsStatsSchema>;
 export type AccountsPartition = z.infer<typeof accountsPartitionSchema>;
 export type PartitionTier = z.infer<typeof partitionTierSchema>;
 export type CompactionState = z.infer<typeof compactionStateSchema>;
+
+export type AggGranularity = z.infer<typeof aggGranularitySchema>;
+export type AggSlots = z.infer<typeof aggSlotsSchema>;
+export type AggRevenue = z.infer<typeof aggRevenueSchema>;
+export type AggShreds = z.infer<typeof aggShredsSchema>;
```
amliu-jumptrading: nit: same here remove slots and shreds

src/api/useSetAtomWsData.ts
```diff
@@ -765,6 +772,27 @@ function useUpdateAtoms() {
...
+              setAggShreds(value);
+              break;
+            }
+            case "query_shreds": {
+              // TODO: handle historical shred updates
+              break;
+            }
+          }
+          break;
+        }
```
amliu-jumptrading: nit: we can remove slots and shreds from here to keep this scoped

src/features/Replay/RevenueTrack/RevenueTrack.tsx
```diff
@@ -0,0 +1,194 @@
...
+  type ExplorableChartProps,
+  type MarkerLinesProps,
+  type RangeChangeSubscriberProps,
+} from "../const.ts";
+import { useThrottledCallback } from "use-debounce";
+import type { WebGlRemountProps } from "../../WebGl/withWebGlRemount.tsx";
+import { useWebGlEventHandlers } from "../../WebGl/useWebGlEventHandlers.ts";
+import withWebGlRemount from "../../WebGl/withWebGlRemount.tsx";
+import {
+  drawAggRevenue as drawAggRevenue,
```
amliu-jumptrading: nit: no need to rename here

src/features/Replay/Chart.tsx
```diff
@@ -0,0 +1,178 @@
...
+    const newWorldEndNs =
+      BigInt(new Date().getTime() - LIVE_CHART_DELAY_MS) * BigInt(nsPerMs);
+    const newWorldEndMs = calcRelativeMs(referenceNs, newWorldEndNs);
+
+    // delay if too soon after startup
+    if (newWorldEndMs < 0) return;
+
+    if (rangeRef.current) {
+      rangeRef.current.worldEndMs = newWorldEndMs;
+      return;
```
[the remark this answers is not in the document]
asuzuki-jumptrading: This will be addressed when we implement live mode

src/features/Replay/useExplorableChart.ts
```diff
@@ -0,0 +1,177 @@
...
+      if (!isZoomingOut && span * scale < MIN_VISIBLE_MS) {
+        scale = MIN_VISIBLE_MS / span;
+      }
+      setVisibleRange([
+        cursorTs - (cursorTs - startTs) * scale,
+        cursorTs + (endTs - cursorTs) * scale,
+      ]);
+    };
+
+    const setUpExploreListeners = (trackEl: HTMLDivElement) => {
```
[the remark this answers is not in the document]
asuzuki-jumptrading: This will be fine tuned later

src/features/Replay/RevenueTrack/RevenueTrack.tsx
amliu-jumptrading: ✂️ no need to rename

src/features/Replay/RevenueTrack/useAggRevenueQuery.ts
```diff
@@ -0,0 +1,39 @@
+import { useCallback } from "react";
+import { ascBucketGranularities, msBucketSizes } from "../const";
+import type { AggGranularity } from "../../../api/types";
+import { useWebSocketSend } from "../../../api/ws/utils";
+import type { NsTsRange } from "../../WebGl/webglUtils";
+
+// TODO: throttle across revenue types
```
amliu-jumptrading, line 7: Rather than calling this hook in `RevenueTrack` and throttling requests (needed once we add the other revenue types), can we let the Chart own the querying and let the Track only handle reading the data from a shared atom and rendering it?

The approach would be:
- Chart computes the granularity once and fires a single request per range change (this eliminates the need to coordinate throttling across per-track requests)
- Update a shared atom with the response (already done in `useSetAtomWsData`). Each revenue type can get the data relevant to itself from the atom
- Tracks subscribe/unsubscribe to queries and the Chart only queries when >= 1 track is subscribed
  - Subscription is tracked by a ref of a set owned by `useAggRevenueQuery` and the hook provides `subscribeAggRevenue` and `unsubscribeAggRevenue` functions that the tracks consume

We'll have more tracks sharing queries in the future, so it would be good to determine which component owns this logic early. I prefer the Chart owning the querying because this lets us reason about it with all/multiple tracks in mind vs scattering the logic over N tracks. It also separates rendering (track) from data (query hook called in the chart).

If we want to support different granularities per track in the future, we can refactor the subscription model to have tracks report their desired granularity then fire one query per distinct granularity and key the shared atom by granularity.

src/features/Replay/Chart.tsx
```diff
@@ -0,0 +1,178 @@
...
+      worldEndMs: newWorldEndMs,
+      visibleRangeMs,
+    };
+    broadcastVisibleRangeChange();
+    refreshSelectedMarkerLine();
+    setIsRangeInitialized(true);
+  }, [
+    currentSlot,
+    startupTimeNs,
+    setVisibleRange,
```
amliu-jumptrading: Is this supposed to be in the dependency array? I don't see it being used

src/features/Replay/RevenueTrack/utils.ts
```diff
@@ -0,0 +1,220 @@
...
+  if (maxValue === 0n) return 0;
+  const ratio = 1 / logRatio(Number(maxValue), Number(value), revenueLogBase);
+  return clampNonZeroValue(ratio, minNonZeroY, maxY);
+}
+
+export function drawAggRevenue(
+  rendererObj: RendererObj,
+  type: RevenueType,
+  aggRevenue: AggRevenue,
+  getRelativeMs: (absoluteMs: bigint) => number,
```
amliu-jumptrading: nit: should this be `absoluteNs` instead?

src/features/Replay/RevenueTrack/utils.ts
```diff
@@ -0,0 +1,220 @@
...
+      return acc;
+    },
+    [],
+  );
+
+  const { cameraReferenceMs, scene, currentMesh, prevMesh } =
+    rendererObj.aggResources;
+  const mesh = prevMesh;
+
+  /** store mesh positions relative to referenceX. This allows CPU to see small coordinates */
```
amliu-jumptrading: nit: should this be GPU?

src/features/Replay/RevenueTrack/utils.ts
```diff
@@ -0,0 +1,220 @@
...
+
+export function setUpAggResources(
+  getWasContextLost: ContextHelpers["getWasContextLost"],
+): AggResources & { cleanUpResources: () => void } {
+  const scene = new THREE.Scene();
+  const camera = new THREE.OrthographicCamera(0, 0, maxY, minY, 0.5, 10);
+  camera.position.z = 1;
+
+  const resources = createWebglResources(REVENUE_OPACITY);
+  const prevMesh = createSlotMesh(resources);
```
amliu-jumptrading: nit: should we rename this to `createRectMesh` since for the revenue tracks, we're not using them as a mesh per slot

src/features/Replay/RevenueTrack/utils.ts
```diff
@@ -0,0 +1,220 @@
...
+  // TODO: add nonAggResources
+  cleanUp: () => void;
+}
+
+export interface AggResources {
+  camera: THREE.OrthographicCamera;
+  scene: THREE.Scene;
+  resources: WebglResources;
+  prevMesh: SlotMesh;
+  currentMesh: SlotMesh;
```
amliu-jumptrading: Do we actually need 2 meshes here? 2 meshes is useful when we want to show one while building the other, but it doesn't seem necessary for this Track. Each draw rebuilds the mesh synchronously and renders in the same tick so we'll never see a partial mesh

src/features/Replay/chart.module.css
```diff
@@ -0,0 +1,18 @@
+.container {
+  padding: 32px;
+  border-radius: 8px;
+  border: 4px solid rgb(42, 53, 72);
+  background: rgb(34, 44, 61);
```
amliu-jumptrading: Do we want to keep this border and background? Designs aren't finalized yet, but they point in the direct of tracks not having a background color and non-rounded edges.

src/features/Replay/RevenueTrack/useAggRevenueQuery.ts
asuzuki-jumptrading: Discussed offline -- we'll keep the individual tracks in charge of triggering the data fetch. The query hooks can keep track of current state / pending requests as needed to prevent duplicates

src/features/Replay/Chart.tsx
asuzuki-jumptrading: good catch! will remove

src/features/Replay/RevenueTrack/utils.ts
asuzuki-jumptrading: yes, thank you!

src/features/Replay/RevenueTrack/utils.ts
asuzuki-jumptrading: yes!

src/features/Replay/RevenueTrack/utils.ts
asuzuki-jumptrading: Discussed offline -- will remove since the current simple strategy is to replace the previous mesh completely

src/features/Replay/chart.module.css
asuzuki-jumptrading: I can remove this. This was based on the initial prototype

src/features/WebGl/webglUtils.ts
```diff
@@ -93,10 +146,9 @@ export function disposeWebglResources(resources: WebglResources) {
 const INITIAL_CAPACITY = 700 * (SHRED_EVENT_TYPES_COUNT - 1);
 
 /**
- * Create one mesh per slot. That way, we can easily add / delete slots without updating
- * unchanged slots.
+ * Create a mesh to draw 2D rectangles
  */
-export function createSlotMesh(resources: WebglResources): SlotMesh {
+export function createRectMesh(resources: WebglResources): SlotMesh {
```
amliu-jumptrading: nit: can we make the type `RectMesh` to map?

src/features/Replay/RevenueTrack/utils.ts
```diff
@@ -0,0 +1,206 @@
...
+    [],
+  );
+
+  const { cameraReferenceMs, mesh } = rendererObj.aggResources;
+
+  /** store mesh positions relative to referenceX. This allows GPU to see small coordinates */
+  if (mesh.referenceX == null) {
+    mesh.referenceX = cameraReferenceMs;
+  }
+  mesh.mesh.position.x = mesh.referenceX - cameraReferenceMs;
```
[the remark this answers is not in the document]
asuzuki-jumptrading: fixed

src/features/Replay/RevenueTrack/RevenueTrack.tsx
```diff
@@ -0,0 +1,192 @@
...
+    rendererRef.current.renderer.setSize(width, height);
+    renderActive();
+  }, [renderActive, width]);
+
+  // trigger draw
+  useLayoutEffect(() => {
+    if (!rendererRef.current || !aggRevenue) return;
+    // TODO: draw non-agg
+    drawAggRevenue(rendererRef.current, type, aggRevenue, getRelativeMs);
+    renderActive();
```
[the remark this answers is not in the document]
asuzuki-jumptrading: The querying logic will be updated with batched fetching. That will be in a separate PR. For now, we can assume responses arrive in request order

src/features/Replay/Chart.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+  const explorableChartProps = useExplorableChart({
+    rangeRef,
+    setSelectedMs,
+    setVisibleRange,
+  });
+
+  // refresh world size
+  useLayoutEffect(() => {
+    const referenceNs = rangeRef.current?.referenceNs ?? startupTimeNs;
```
atsai-jumptrading, line 130: this is always `startupTimeNs` it's just circular. startupTimeNs -> const -> ref -> const

src/features/Replay/Chart.tsx
```diff
@@ -0,0 +1,181 @@
...
+    rangeRef.current = {
+      referenceNs,
+      worldEndMs: newWorldEndMs,
+      visibleRangeMs,
+    };
+    broadcastVisibleRangeChange();
+    refreshSelectedMarkerLine();
+    setIsRangeInitialized(true);
+  }, [
+    currentSlot,
```
atsai-jumptrading, line 157: i don't think we want `currentSlot` that'll cause a lot of renders. And we're just using it to tick of current time? We should either subscribe to the atom or have the atom update trigger a tick, either direction. But not have to go through react here

src/features/Replay/Chart.tsx
```diff
@@ -0,0 +1,181 @@
+import { Flex } from "@radix-ui/themes";
+import { useAtomValue } from "jotai";
+import { useRef, useLayoutEffect, useMemo, useCallback, useState } from "react";
+import { useMeasure } from "react-use";
+import styles from "./chart.module.css";
+import { type MarkerLinesProps } from "./const.ts";
+import { nsPerMs } from "../../consts.ts";
+import { clamp } from "../../uplotReact/utils.ts";
+import { calcRelativeMs, getInitVisibleRange } from "./utils.ts";
+import VisibleRange from "./VisibleRangeInfo.tsx";
```
atsai-jumptrading, line 10: Let's call them the same, i was trying to find the `VisibleRange` component but we re-name here

src/features/Replay/Chart.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+/**
+ * Set up Replay chart, which keeps track of a reference ts (startup time), and
+ * visible and world ts ranges.
+ * Informs subscribers of visible range changes.
+ */
+export default function Chart({ startupTimeNs }: ChartProps) {
+  const currentSlot = useAtomValue(currentSlotAtom);
+
+  const [measureRef, { width }] = useMeasure<HTMLDivElement>();
```
atsai-jumptrading: can width be moved down to `RevenueTrack`? Since that is actually what uses the width. It's just used as a guard here and that seems off

src/features/Replay/Chart.tsx
```diff
@@ -0,0 +1,181 @@
...
+   */
+  startupTimeNs: bigint;
+}
+
+/**
+ * Set up Replay chart, which keeps track of a reference ts (startup time), and
+ * visible and world ts ranges.
+ * Informs subscribers of visible range changes.
+ */
+export default function Chart({ startupTimeNs }: ChartProps) {
```
atsai-jumptrading, line 36: the init with the effects is a bit confusing and hard to follow. It's either initialized or not, and that can be determined on render. Then the subsequent updating can be done on a subscription. Then we're guarded on width and the range init, but shouldn't the guard just be on range values?

src/features/Replay/Chart.tsx
```diff
@@ -0,0 +1,181 @@
...
+  const explorableChartProps = useExplorableChart({
+    rangeRef,
+    setSelectedMs,
+    setVisibleRange,
+  });
+
+  // refresh world size
+  useLayoutEffect(() => {
+    const referenceNs = rangeRef.current?.referenceNs ?? startupTimeNs;
+    const newWorldEndNs =
```
atsai-jumptrading, line 131: should this be our adjusted server time like shreds?

src/features/Replay/RevenueTrack/utils.ts
```diff
@@ -0,0 +1,204 @@
...
+    aggResources: omit(aggResources, "cleanUpResources"),
+    cleanUp,
+  };
+}
+
+export function setUpAggResources(
+  getWasContextLost: ContextHelpers["getWasContextLost"],
+): AggResources & { cleanUpResources: () => void } {
+  const scene = new THREE.Scene();
+  const camera = new THREE.OrthographicCamera(0, 0, maxY, minY, 0.5, 10);
```
atsai-jumptrading, line 91: should this be init from (0,1 vs (0,0? Apparently can cause some downstream divide by 0 on init

src/features/Replay/useVisibleRangeSubscribers.ts
```diff
@@ -0,0 +1,100 @@
...
+      } of visibleRangeSubscribersRef.current.values()) {
+        onRangeChange(
+          rangeRef.current.visibleRangeMs,
+          [0, rangeRef.current.worldEndMs],
+          selectedMsRef.current,
+        );
+      }
+    };
+
+    const broadcastSelectedMsChange = () => {
```
atsai-jumptrading, line 49: We've separated out the editing of the values like `worldEndMs` and the broadcast to 2 separate actions. Is that intentional that we want them to be separate? The potential issues is we update the value and don't broadcast. Jotai does also do this. Also could make it granular where visibleRange isn't coupled to worldEndMs  though not sure if that's an issue if it's all in one

src/features/Replay/useExplorableChart.ts
```diff
@@ -0,0 +1,190 @@
...
+          e.preventDefault();
+        },
+        onTouchMove: (e: TouchEvent) => {
+          if (e.touches.length !== 1) return;
+          moveDrag(e.touches[0].clientX);
+          e.preventDefault();
+        },
+        onWheel: (e: WheelEvent) => {
+          e.preventDefault();
+          zoom(e.clientX, e.deltaY);
```
atsai-jumptrading, line 149: apparently we should be checking `deltaMode` for browsers

src/features/Replay/useExplorableChart.ts
```diff
@@ -0,0 +1,190 @@
...
+        onMouseMove,
+        onTouchStart,
+        onTouchMove,
+        onWheel,
+      } = createCallbacks(trackEl, refreshCursor, false);
+
+      const cleanups = [
+        addListener(trackEl, "mousedown", onMouseDown),
+        addListener(trackEl, "mousemove", onMouseMove),
+        addListener(trackEl, "mouseup", endDrag),
```
atsai-jumptrading, line 177: i believe we'd lose this if we mouse up'd outside the track? Youse use mouseleave i think. Does setPointerCapture help? Any multiple pointer issues once we have more tracks?

src/features/Replay/useExplorableChart.ts
```diff
@@ -0,0 +1,190 @@
...
+        },
+      };
+    },
+    [rangeRef, setSelectedMs, setVisibleRange],
+  );
+
+  return useMemo(() => {
+    const setUpExploreListeners = (trackEl: HTMLDivElement) => {
+      const refreshCursor = () => {
+        const cursor = isPanningRef.current ? "grabbing" : "grab";
```
atsai-jumptrading, line 159: can we use css `:active`?

src/features/Replay/useExplorableChart.ts
```diff
@@ -0,0 +1,190 @@
...
+      refreshCursor();
+
+      const {
+        endDrag,
+        onMouseDown,
+        onMouseMove,
+        onTouchStart,
+        onTouchMove,
+        onWheel,
+      } = createCallbacks(trackEl, refreshCursor, false);
```
atsai-jumptrading, line 172: callback within memo is necessary? I didn't try to re-arrange everything but feels off

src/features/Replay/RevenueTrack/RevenueTrack.tsx
```diff
@@ -0,0 +1,192 @@
...
+  const subscriptionId = `${type}-${baseSubscriptionId}`;
+  const [granularity, setGranularity] = useState<AggGranularity | undefined>(
+    undefined,
+  );
+
+  const containerRef = useRef<HTMLDivElement>(null);
+  const rendererRef = useRef<RendererObj | undefined>();
+
+  const widthRef = useRef(width);
+  widthRef.current = width;
```
atsai-jumptrading, line 56: do we need this? We use this for initial sizing, but also have an effect that sets the size right after anyways. Does `setupRendereres` even need a width?

atsai-jumptrading (approved):

src/features/Replay/Chart.tsx
asuzuki-jumptrading: yeah let me clean up how it gets initialized

src/features/Replay/Chart.tsx
asuzuki-jumptrading: yeah we can do that

src/features/Replay/useVisibleRangeSubscribers.ts
asuzuki-jumptrading: The broadcasting is intentionally separated. The selected ts can change separately from the visible range. The world range can also change separately from those two. The way I planned to use it is:
- if visible range changes (along with other two changing), `broadcastVisibleRangeChange`
- else if selected range changes, `broadcastSelectedMsChange`
- else if world range changes, broadcast world range change (this is in the next mini map PR)

src/features/Replay/useExplorableChart.ts
asuzuki-jumptrading: Currently the drag ends when the mouse leaves the track.
There will be another PR where we move the mouse move and mouse up to the tracks container, instead of individual tracks

src/features/Replay/useVisibleRangeSubscribers.ts
asuzuki-jumptrading: Discussed offline -- we're going to switch to using atoms for visible, world and selected. It'll be simpler :)

src/features/Replay/useExplorableChart.ts
asuzuki-jumptrading: I'm going to leave this as is for now. We might be able to move to :active on the container element instead of each individual track.
For mini maps (next PR), we want the resizing cursor as well

src/features/Replay/useExplorableChart.ts
asuzuki-jumptrading: Yes, this is in anticipation of the Mini Map (next PR) which will reuse some fns but for the mini map track el 😅

src/features/Replay/RevenueTrack/RevenueTrack.tsx
asuzuki-jumptrading: Let me see if I can remove this along with the other refactors. The main issue is that the initial resizing could happen before setupRenderers fires, in which case `setupRenderers` needs to know the initial width. I needed the ref to prevent the unmount code from triggering on width change.

src/features/Replay/Chart.tsx
asuzuki-jumptrading: I'll remove `isRangeInitialized` and have each track handle the state when it has no range info

src/features/Replay/RevenueTrack/RevenueTrack.tsx
asuzuki-jumptrading: I removed the widthRef but needed to add an isInitialized state to ensure that the width is corrected after setup

src/features/Replay/Chart.tsx
asuzuki-jumptrading: This is now a raf loop

src/features/Replay/Chart.tsx
asuzuki-jumptrading: Will use the atom from this PR: https://github.com/firedancer-io/firedancer-frontend/pull/436

src/features/Replay/index.tsx
```diff
@@ -1,19 +1,29 @@
...
-          <Flex direction="column" gap="4" mt="2">
-            Replay Chart
-          </Flex>
-        </Card>
+        startupTimeNs != null ? (
+          <Card>
+            <CardHeader text="Replay" />
+            <Flex direction="column" gap="4" mt="2">
+              <Chart
+                key={String(startupTimeNs)}
```
atsai-jumptrading: mismatch with the atoms now since the atoms aren't reset with this

src/features/Replay/atoms.ts
```diff
@@ -0,0 +1,34 @@
+import type { TsRange } from "../WebGl/webglUtils";
+import { atom } from "jotai";
+
+export const { referenceNsAtom, visibleRangeAtom, worldRangeAtom } =
+  (function getReplayChartAtoms() {
+    const _referenceNsAtom = atom<bigint | undefined>();
+    const _visibleRangeAtom = atom<TsRange | undefined>(undefined);
```
atsai-jumptrading, line 7: you used to do an check to only update if changed, might still want that

atsai-jumptrading (approved):

src/features/Replay/Chart.tsx
```diff
@@ -0,0 +1,141 @@
...
+  // refresh world size
+  useRafLoop((time: number) => {
+    if (time - lastWorldUpdateTsRef.current < WORLD_UPDATE_INTERVAL_MS) return;
+    lastWorldUpdateTsRef.current = time;
+
+    const newWorldEndNs =
+      BigInt(new Date().getTime() - LIVE_CHART_DELAY_MS) * BigInt(nsPerMs);
+    const referenceNs = store.get(referenceNsAtom);
+
+    if (referenceNs != null) {
```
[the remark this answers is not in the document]
asuzuki-jumptrading: We decided that we're going to clear the entire app state when validator restarts, so this will no longer be an issue

atsai-jumptrading (approved):

src/features/WebGl/webglUtils.ts
```diff
@@ -128,67 +180,68 @@ export function createSlotMesh(resources: WebglResources): SlotMesh {
     colorAttr,
     capacity: INITIAL_CAPACITY,
     count: 0,
+    referenceX: undefined,
   };
 }
 
 /**
  * Bump up capacity for slot as needed
```
amliu-jumptrading: nit: `slot` should be `rect` here

amliu-jumptrading (approved):

PR #413, opened 2026-08-26
feat: mini map

asuzuki-jumptrading:
- mini map using aggregated slots data
- drag visible range box to pan
- scroll or drag left / right handles to zoom


src/features/Replay/MiniMap/MiniMap.tsx
```diff
@@ -0,0 +1,201 @@
...
+    >
+      <div
+        ref={chartContainerRef}
+        className={clsx(styles.miniMapTrack, miniMapMarkerLinesClassName)}
+      />
+      <div ref={visibleRangeElRef} className={styles.visibleRangeBox}>
+        <div ref={leftHandleRef} className={clsx(styles.handle, styles.left)} />
+        <div
+          ref={rightHandleRef}
+          className={clsx(styles.handle, styles.right)}
```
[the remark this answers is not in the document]
asuzuki-jumptrading: We will be updating pan / zoom controls in a later PR

src/features/Replay/MiniMap/miniMap.module.css
```diff
@@ -0,0 +1,50 @@
...
+    &::before {
+      content: "";
+      position: absolute;
+      /* expand clickable range */
+      inset: -5px -10px;
+    }
+  }
+
+  .handle {
+    display: none;
```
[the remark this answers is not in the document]
asuzuki-jumptrading: We'll deal with these controls in a later PR

src/features/Replay/useExplorableChart.ts
```diff
@@ -205,6 +216,170 @@ export function useExplorableChart(): ExplorableChartProps {
...
+      refreshCursor();
+
+      const {
+        endDrag,
+        onMouseDown,
+        onMouseMove,
+        onTouchStart,
+        onTouchMove,
+        onWheel,
+      } = createCallbacks(trackEl, refreshCursor, true);
```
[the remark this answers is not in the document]
asuzuki-jumptrading: fixed

src/features/Replay/Chart.tsx
```diff
@@ -55,15 +60,23 @@ export default function Chart() {
...
+    if (!visibleRange || !worldRange) return;
     const [start, end] = visibleRange;
     const pct = (100 * (selectedMs - start)) / (end - start);
     containerRef.current.style.setProperty(MARKER_PCT_VAR, `${pct}%`);
+
+    const worldPct = (100 * selectedMs) / (worldRange[1] - worldRange[0]);
+    containerRef.current.style.setProperty(
+      WORLD_MARKER_PCT_VAR,
+      `${worldPct}%`,
+    );
```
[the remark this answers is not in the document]
asuzuki-jumptrading: fixed

src/features/Replay/MiniMap/useMiniMapQuery.ts
```diff
@@ -0,0 +1,64 @@
...
+          query([lastRequest.worldRangeNs[1], worldRangeNs[1]], granularity);
+        } else {
+          // fetch entire world on granularity change or on start range change
+          query(worldRangeNs, granularity);
+        }
+
+        lastRequestRef.current = {
+          worldRangeNs,
+          granularity,
+        };
```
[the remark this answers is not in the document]
asuzuki-jumptrading: We'll add a broader strategy later for handling disconnects

src/features/Replay/Chart.tsx
```diff
@@ -49,7 +54,26 @@ export default function Chart() {
...
+    if (selectedMs == null) {
+      // off screen
+      containerRef.current.style.setProperty(WORLD_MARKER_PCT_VAR, "-300%");
+      return;
+    }
+
+    const worldRange = store.get(worldRangeAtom);
+    if (!worldRange) return;
+
+    const worldPct = (100 * selectedMs) / (worldRange[1] - worldRange[0]);
```
amliu-jumptrading: For the visible range, we're calculating the `pct` as
```
const [start, end] = visibleRange;
const pct = (100 * (selectedMs - start)) / (end - start);
```
This works fine because `worldRange[0] = 0`, but we change this so it doesn't rely on this assumption.
```
const [start, end] = worldRange;
const pct = (100 * (selectedMs - start)) / (end - start);
```


On related note, the `refreshWorldRangeSelectedMarker` and `refreshVisibleRangeSelectedMarker` functions are remarkably similar, only differing in the range atom and marker pct css var. We can make a factory function to prevent similar implementation drifts between the visible and world ranges refreshers.
```
const createRefreshSelectedMarker = useCallback(
   (rangeAtom: Atom<TsRange | undefined>, cssVar: string) => () => {
      ...
   },
   []
);

useLayoutEffect(() => {
   const refreshWorldRangeSelectedMarker = createRefreshSelectedMarker(
      worldRangeAtom,
      WORLD_MARKER_PCT_VAR,
    );
    const refreshVisibleRangeSelectedMarker = createRefreshSelectedMarker(
      visibleRangeAtom,
      MARKER_PCT_VAR,
    );
   ....
}, [createRefreshSelectedMarker]);
```

src/api/useSetAtomWsData.ts
```diff
@@ -811,6 +811,10 @@ function useUpdateAtoms() {
         }
         case "timeline": {
           switch (key) {
+            case "query_agg_slots": {
```
amliu-jumptrading: nit: mind moving this case to the bottom of the timeline switch? W usually put the cases that are just there for type completeness (see the summary switch) after the ones that update an atom

src/features/Replay/MiniMap/MiniMap.tsx
```diff
@@ -0,0 +1,201 @@
...
+      trackHeight,
+      setUpContextListeners,
+      getWasContextLost,
+    );
+    if (!rendererObj) return;
+
+    rendererRef.current = rendererObj;
+    chartContainerRef.current.replaceChildren(rendererObj.renderer.domElement);
+
+    const unsusbscribes = [
```
amliu-jumptrading: nit: typo `unsubscribes`

src/features/Replay/MiniMap/MiniMap.tsx
```diff
@@ -0,0 +1,201 @@
...
+      moveCamera(rendererRef.current, worldRangeMs);
+      render(rendererRef.current);
+    },
+    [query],
+  );
+
+  /**
+   * Update camera and query data for new range
+   */
+  const onRangeChange = useCallback(() => {
```
amliu-jumptrading: Could we rename this to `onWorldRangeChange` and add a separate `onVisibleRangeChange` that only calls `updateVisibleEl`? Right now `visibleRangeAtom` and `worldRangeAtom` share a listener, so every `visibleRangeAtom` update also calls `updateWorldEl`, which re-renders the scene unnecessarily (the mini-map scene only depends on the world range).

src/features/Replay/MiniMap/utils.ts
```diff
@@ -0,0 +1,279 @@
...
+  referenceMs: number;
+}
+
+export type RendererObj = {
+  renderer: THREE.WebGLRenderer;
+  camera: THREE.OrthographicCamera;
+  scene: THREE.Scene;
+  /* resources shared by this renderer's meshes */
+  resources: WebglResources;
+  meshes: RectMesh[];
```
amliu-jumptrading: In practice, we'll almost never need multiple meshes because `getMiniMapGranularity` limits us to `BUCKET_COUNT_THRESHOLD` buckets until we run out of coarser granularities (1d currently). Based on the current values (`MAX_RECTANGLES_PER_MESH = 8000`, 2 states i.e. 2 rects per bucket, and 1d granularity), we'll only ever use a 2nd mesh if we have ~11 yrs worth of data.
While the 11 yr calculation will change if adjust any of those values, the GPU should be able to handle orders of magnitude more rects than we'd ever produce here.

Could we simplify to a single mesh that grows via `ensureCapacity` and drop the multi-mesh index math?

src/features/Replay/MiniMap/utils.ts
```diff
@@ -0,0 +1,279 @@
...
+    MAX_WEBGL_PX_RATIO,
+    setUpContextListeners,
+    getWasContextLost,
+  );
+  if (!rendererObj) return;
+
+  const { renderer, cleanUpRenderer } = rendererObj;
+
+  const scene = new THREE.Scene();
+  const camera = new THREE.OrthographicCamera(0, 0, maxY, minY, 0.5, 10);
```
amliu-jumptrading: Do we need `const camera = new THREE.OrthographicCamera(0, 1, maxY, minY, 0.5, 10);` here? We changed it in the revenue track in the last PR.

asuzuki-jumptrading: ah yes good point

src/features/Replay/MiniMap/utils.ts
asuzuki-jumptrading: Yeah, let me remove this as it will most likely not be necessary even if we made the mini map an exception with much more granular data. Thanks for checking!

src/features/Replay/Chart.tsx
asuzuki-jumptrading: Thanks for catching this! Now that I'm storing worldRange instead of worldEnd, it's good to add the - world range start in these places

PR #419, opened 2026-08-31
feat: slot timeline lanes

asuzuki-jumptrading:


src/features/Overview/SlotTimeline/useSlotLanes.ts
```diff
@@ -0,0 +1,99 @@
...
+
+const nonAlpenglowSlotLanesAtom = atom((get) => {
+  const storageSlot = get(storageSlotAtom);
+  const rootSlot = get(rootSlotAtom);
+  const voteSlot = get(voteSlotAtom);
+  const repairSlot = get(repairSlotAtom);
+  const turbineSlot = get(turbineSlotAtom);
+  const replaySlot = get(completedSlotAtom);
+  const optimisticallyConfirmedSlot = get(optimisticallyConfirmedSlotAtom);
+  const nextLeaderSlot = get(nextLeaderSlotAtom);
```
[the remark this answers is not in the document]
asuzuki-jumptrading: fixed

src/features/Overview/SlotTimeline/SlotLanes.tsx
```diff
@@ -0,0 +1,292 @@
...
+
+  if (leftRange == null) return null;
+
+  const leftBarsCount = leftRange.maxSlot - leftRange.minSlot + 1;
+  const nextLeaderLane = lanes[nextLeaderLaneIdx];
+  const hasNextLeader = lanes[nextLeaderLaneIdx].slot != null;
+  const nextSlotsCount =
+    nextLeaderLane.slot == null
+      ? undefined
+      : nextLeaderLane.slot - leftRange.maxSlot;
```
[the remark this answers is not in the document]
asuzuki-jumptrading: fixed, with min set to 1

src/features/Overview/SlotTimeline/SlotLanes.tsx
```diff
@@ -0,0 +1,293 @@
...
+    if (!range) {
+      return undefined;
+    }
+
+    return {
+      maxSlot: range.max,
+      minSlot: Math.max(
+        0,
+        Math.min(range.max - minLeftSlotsCellsCount + 1, range.min),
+      ),
```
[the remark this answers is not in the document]
asuzuki-jumptrading: fixed

src/features/Overview/SlotTimeline/utils.ts
```diff
@@ -0,0 +1,109 @@
...
+  const nextLeaderWidth = hasNextLeader ? widths.slotCell : "";
+  const columns = `repeat(${leftSlotCellsCount}, ${widths.slotCell}) ${widths.nextSlots} ${nextLeaderWidth}`;
+
+  return {
+    columns,
+    barsGap: gap,
+  };
+}
+
+function getSlotCellsNeededWith(slotCellsCount: number, barsGap: number) {
```
amliu-jumptrading: nit: typo `getSlotCellsNeededWidth`

src/features/Overview/SlotTimeline/utils.ts
```diff
@@ -0,0 +1,109 @@
...
+  if (defaultNeededWidth <= defaultAvailableWidth) {
+    return { gap: defaultBarsGap, useMinNextBarsWidth: false };
+  }
+
+  // collapse next bars to min width instead of width %
+  const maxAvailableWidth = containerWidth - nextBarsBoxMinWidth;
+
+  for (const gap of [defaultBarsGap, smallBarsGap, 0]) {
+    const needed = getSlotCellsNeededWith(slotCellsCount, gap);
+    if (needed < maxAvailableWidth) {
```
amliu-jumptrading: Should this be `<=` like above: `defaultNeededWidth <= defaultAvailableWidth`

src/features/Overview/SlotTimeline/utils.ts
```diff
@@ -0,0 +1,109 @@
...
+  };
+}
+
+export function getGridColumnsAndGap(
+  leftSlotCellsCount: number,
+  hasNextLeader: boolean,
+  containerWidth: number,
+) {
+  const slotCellsCount = leftSlotCellsCount + (hasNextLeader ? 1 : 0);
+  const { gap, useMinNextBarsWidth } = getSlotCellsGap(
```
amliu-jumptrading: nit: `useMinNextBarsWidth` reads like a hook name. Can we perhaps use `pinNextBarsToMin`?

src/features/Overview/SlotTimeline/utils.ts
```diff
@@ -0,0 +1,109 @@
...
+    defaultBarsGap,
+  );
+  if (defaultNeededWidth <= defaultAvailableWidth) {
+    return { gap: defaultBarsGap, useMinNextBarsWidth: false };
+  }
+
+  // collapse next bars to min width instead of width %
+  const maxAvailableWidth = containerWidth - nextBarsBoxMinWidth;
+
+  for (const gap of [defaultBarsGap, smallBarsGap, 0]) {
```
amliu-jumptrading: We can drop gap = 0 from the loop because its return value is equivalent to the fallback return value

src/features/Overview/SlotTimeline/useSlotLanes.ts
```diff
@@ -0,0 +1,100 @@
...
+    pinnedLaneIdx: lanes.indexOf(replaySlotBar),
+  };
+});
+
+export function useSlotLanes(): {
+  lanes: SlotLaneInfo[];
+  nextLeaderLaneIdx: number;
+  pinnedLaneIdx: number;
+} {
+  return useAtomValue(nonAlpenglowSlotLanesAtom);
```
amliu-jumptrading: Are we supposed to have an `alpenglowSlotLanesAtom` as well?

src/features/Overview/SlotTimeline/SlotLanes.tsx
```diff
@@ -0,0 +1,307 @@
...
+        <Text className={styles.nextLeaderTime} truncate dir="rtl">
+          {nextLeaderSlot == null ? "∞" : nextSlotText}
+        </Text>
+      </Flex>
+      <div>
+        <Progress
+          value={progressSinceLastLeader}
+          height="1px"
+          // a bit longer than an expected slot duration
+          duration="500ms"
```
amliu-jumptrading: Should we use `target_slot_duration_nanos` from the epoch message here so it responds to shorter slot durations?

src/features/Overview/SlotTimeline/SlotLanes.tsx
```diff
@@ -0,0 +1,307 @@
...
+        range.max - minLeftSlotsCellsCount + 1,
+      ),
+    };
+  }, [lanes, nextLeaderLaneIdx]);
+
+  if (leftRange == null) return null;
+
+  const leftBarsCount = leftRange.maxSlot - leftRange.minSlot + 1;
+  const nextLeaderLane = lanes[nextLeaderLaneIdx];
+  const hasNextLeader = lanes[nextLeaderLaneIdx].slot != null;
```
amliu-jumptrading: nit: we can do `nextLeaderLane.slot != null` here

src/features/Overview/SlotTimeline/slotLanes.module.css
```diff
@@ -0,0 +1,147 @@
...
+  }
+  .slot-dt {
+    font-weight: 600;
+    min-width: 20px;
+  }
+}
+
+.slot-lane-bars {
+  display: contents;
+  overflow: hidden;
```
amliu-jumptrading: nit: overflow hidden doesn't have any effect on a container with display contents so we can remove it

src/features/Overview/SlotTimeline/utils.ts
asuzuki-jumptrading: ah yes

src/features/Overview/SlotTimeline/useSlotLanes.ts
asuzuki-jumptrading: Yes, in a separate PR

asuzuki-jumptrading: I didn't want to block this non-alpenglow release for the alpenglow data

src/features/Overview/SlotTimeline/SlotLanes.tsx
asuzuki-jumptrading: Yes, we can do that. This was copied from the existing countdown component

src/features/Overview/SlotTimeline/utils.ts
asuzuki-jumptrading: Let me just type these out explicitly since there are only 3 and one is the fallback

src/features/Overview/SlotTimeline/utils.ts
amliu-jumptrading: Can we rename `useMinNextBarsWidth` to `pinNextBarsToMin` everywhere it's used, not just the prop for `getGridColumnWidths` 😅

amliu-jumptrading (dismissed):

src/features/Overview/SlotTimeline/utils.ts
asuzuki-jumptrading: oh yes definitely 😅 sorry I missed the other places!

amliu-jumptrading (dismissed):

src/features/Overview/SlotTimeline/SlotLanes.tsx
```diff
@@ -0,0 +1,308 @@
...
+    </Flex>
+  );
+});
+
+interface NextSlotsProps {
+  count?: number;
+}
+
+function NextSlots({ count }: NextSlotsProps) {
+  const [measureRef, { width }] = useMeasure<HTMLDivElement>();
```
atsai-jumptrading: here too?

src/features/Overview/SlotTimeline/SlotLanes.tsx
```diff
@@ -0,0 +1,308 @@
...
+          const highlightedIdx =
+            lane.slot == null ? undefined : lane.slot - leftRange.minSlot;
+          return (
+            <MSlotLaneBars
+              key={lane.label}
+              className={lane.className}
+              leftBarsCount={leftBarsCount}
+              highlightedIdx={highlightedIdx}
+              hasNextLeader={hasNextLeader}
+              highlightNextLeaderCell={hasNextLeader && isNextLeaderLane}
```
atsai-jumptrading: it's kinda weird having 2 different pieces of logic that control the highlighting, and you end up with the potential to have 2 highlights because of that since nothing makes them mutually exclusive. I know currently logic wise it can't but because of the slot math but could cut the highlight idx into 1 prop 
```
const highlightedIdx =
    lane.slot == null
      ? undefined
      : lane.isNextLeader
        ? leftBarsCount
        : lane.slot - leftRange.minSlot;
```

src/features/Overview/SlotTimeline/SlotLanes.tsx
```diff
@@ -0,0 +1,308 @@
...
+
+    if (!range) {
+      return undefined;
+    }
+
+    return {
+      maxSlot: range.max,
+      minSlot: Math.max(range.min, range.max - maxLeftSlotsCellsCount + 1),
+    };
+  }, [lanes, nextLeaderLaneIdx]);
```
atsai-jumptrading: the memo doesn't really do anything since lanes is reallocated every time, but also both these deps are from `useSlotLanes` so why not just move the logic to the hook instead of the component. Then instead of the reduce you also make the skip leader lane logic explicit. Something like this all within the `useSlotLanes` hook
```
export function getLeftRange(lanes: SlotLaneInfo[]) {
  const slots = lanes.map(({ slot }) => slot).filter((slot) => slot != null);
  if (!slots.length) return undefined;

  const maxSlot = Math.max(...slots);
  return {
    maxSlot,
    minSlot: Math.max(Math.min(...slots), maxSlot - maxLeftSlotsCellsCount + 1),
  };
}

const leftLanes = [
  turbineSlotBar, repairSlotBar, replaySlotBar,
  optimisticallyConfirmedBar, voteSlotBar, rootSlotBar, storageSlotBar,
];

const lanes = [nextLeaderSlotBar, ...leftLanes];
const leftRange = getLeftRange(leftLanes),
```

src/features/Overview/SlotTimeline/useSlotLanes.ts
```diff
@@ -0,0 +1,100 @@
...
+    replaySlotBar,
+    optimisticallyConfirmedBar,
+    voteSlotBar,
+    rootSlotBar,
+    storageSlotBar,
+  ];
+
+  return {
+    lanes,
+    nextLeaderLaneIdx: lanes.indexOf(nextLeaderSlotBar),
```
atsai-jumptrading: simpler to just put these idx on `SlotLaneInfo` i think. It's used for the plus sign or icon and that's where you ahve the slot info anyways then you're not having to pass this stuff separtely

src/features/Overview/SlotTimeline/SlotLanes.tsx
```diff
@@ -0,0 +1,308 @@
...
+      {Array.from({ length: leftBarsCount }, (_, i) => {
+        return (
+          <MCell
+            key={i}
+            isHighlighted={i === highlightedIdx}
+            isTransparent={noGap}
+          />
+        );
+      })}
+      {hasNextLeader && (
```
atsai-jumptrading: if we take out `hasNextLeader` then we also don't need to special case this. The component becomes leader cell agnostic and just does render n cells, and highlight one of them if index match

src/features/Overview/SlotTimeline/SlotLanes.tsx
```diff
@@ -0,0 +1,308 @@
...
+        ref={measureRef}
+      >
+        <div
+          style={{
+            gridColumn: leftBarsCount + 1,
+            gridRowStart: 1,
+            gridRowEnd: lanes.length + 1,
+          }}
+        >
+          <NextSlots count={nextSlotsCount} />
```
atsai-jumptrading: if we have a lot of renders inter slot could consider memo'ing it but i'm not sure if we do

src/features/Overview/SlotTimeline/SlotLanes.tsx
```diff
@@ -0,0 +1,308 @@
...
+        </div>
+
+        {lanes.map((lane, i) => {
+          const isNextLeaderLane = i === nextLeaderLaneIdx;
+          const highlightedIdx =
+            lane.slot == null ? undefined : lane.slot - leftRange.minSlot;
```
atsai-jumptrading: If we go over 500 we just drop the highlight? 

if we go over the max `maxLeftSlotsCellsCount` say min is 1 max is 1000. So get `maxSlot` of 1000 and `minSlot` of 501 with `minSlot = maxSlot - 499`. lane.slot of 1 - 501 = -500? 

and at 500 i'd question if there are other render issues or cells get too small by then to really see what's going on

src/features/Overview/SlotTimeline/SlotLanes.tsx
```diff
@@ -0,0 +1,308 @@
...
+  const nextSlotsCount =
+    nextLeaderLane.slot == null
+      ? undefined
+      : // always show at least one bar
+        Math.max(1, nextLeaderLane.slot - leftRange.maxSlot - 1);
+
+  const { columns, barsGap } = getGridColumnsAndGap(
+    leftBarsCount,
+    hasNextLeader,
+    barsContainerWidth,
```
atsai-jumptrading, line 41: do we have on mount flicker? There's logic of gaps, etc that's dependent on this value which is 0 on mount. Might want to either have 0 be unset or seed a plausible width if so

atsai-jumptrading (dismissed):

src/features/Overview/SlotTimeline/SlotLanes.tsx
asuzuki-jumptrading: Yeah, good idea! I can combine them further by passing in a total count of cells

src/features/Overview/SlotTimeline/SlotLanes.tsx
asuzuki-jumptrading: good point, I'll use the default gap if the container width is 0

src/features/Overview/SlotTimeline/SlotLanes.tsx
asuzuki-jumptrading: Discussed offline: we'll keep this for now and then have a design discussion for outliers handling

src/features/Overview/SlotTimeline/SlotLanes.tsx
asuzuki-jumptrading: Makes sense, thanks! I'll also add the gradual shrinking refs in that hook

amliu-jumptrading (dismissed):

src/features/Overview/SlotTimeline/SlotLanes.tsx
```diff
@@ -0,0 +1,272 @@
...
+  return (
+    <Flex>
+      <Grid
+        className={styles.grid}
+        flexShrink="0"
+        columns="repeat(3, max-content)"
+        gapX="5px"
+        mr="5px"
+      >
+        {lanes.map((lane, i) => {
```
atsai-jumptrading: can remove i now

src/features/Overview/SlotTimeline/utils.ts
```diff
@@ -0,0 +1,120 @@
...
+  return dtSlot - referenceSlot;
+}
+
+export function getSlotLaneInfo(
+  label: string,
+  dtSlot: number | null | undefined,
+  referenceSlot: number | null | undefined,
+  className: string,
+  isPinned: boolean,
+  isNextLeader: boolean,
```
atsai-jumptrading: nit but might be more readable as an options param now so you don't just have repeated booleans that could mistakenly get mixed up

atsai-jumptrading (approved):

PR #422, opened 2026-08-31
chore: alpenglow PeerUpdateVoteAccount response

amliu-jumptrading:
WS Changes
| epoch_credits   | `number`       | *Tower Only* The number of credits earned by the vote account during the current epoch |
| epoch_rewards   | `string`       | *Alpenglow Only* Voting rewards credited to this account during the current epoch, in lamports |

epoch_credits is not currently used anywhere so we can simply remove it to avoid any zod schema validation errors


amliu-jumptrading:
WS message iterated on and this field is no longer touched

PR #425, opened 2026-09-02
omit tpu waterfall, schedule for RPC nodes

ripatel-fd:
RPC nodes ([layout].enable_block_production = false) do not have
leader slots or block production tiles, so skip rendering empty
views.


asuzuki-jumptrading:
Will iterate on this branch

PR #433, opened 2026-09-14
fix: improve slot details performance

asuzuki-jumptrading:
Fixes to make slot details page faster when txns count is large.
- Build a map of bundle states per microblock, so we don't loop through all txns for each txn idx.
- Consolidate txn data collection in Provider, to eliminate looping through all txns in multiple components.


atsai-jumptrading (dismissed):

src/features/SlotDetails/SlotTransactionsProvider.tsx
```diff
@@ -0,0 +1,189 @@
...
+      // collect all txnIdxs
+      const mbId = transactions.txn_microblock_id[i];
+      const txnIdxs = bundleTxnIdxsByMicroblockId.get(mbId) ?? [];
+      txnIdxs.push(i);
+      bundleTxnIdxsByMicroblockId.set(mbId, txnIdxs);
+    }
+
+    const getTxnBundleStats = (txnIdx: number): TxnBundleStats => {
+      const mbId = transactions.txn_microblock_id[txnIdx];
+      const bundleTxnIdx = bundleTxnIdxsByMicroblockId.get(mbId) ?? [];
```
atsai-jumptrading: might wanna rename this one `bundleTxnIdxs` since we're using `txnIdxs` for plural already

asuzuki-jumptrading: ok I'll rename it in other files too for consistency

atsai-jumptrading (dismissed):

src/features/Overview/SlotPerformance/TransactionBarsCard/ChartTooltip.tsx
```diff
@@ -27,28 +25,26 @@ import {
...
 
   const bundleStats = useMemo(() => {
-    if (!transactions) return;
+    if (!transactions || !getTxnBundleStats) return;
     if (txnIdx < 0) return;
     if (!transactions.txn_from_bundle[txnIdx]) return;
 
-    return getTxnBundleStats(transactions, txnIdx);
-  }, [transactions, txnIdx]);
+    return getTxnBundleStats(txnIdx);
```
amliu-jumptrading, line 46: Can we add this guard to `getTxnBundleStats`:
```
if (!transactions.txn_from_bundle[txnIdx]) return;
```

If we do, we can remove the guards here and simplify to:
```
const bundleStats = useMemo(() => getTxnBundleStats?.(txnIdx), [txnIdx, getTxnBundleStats]);
```

We can also simplify the check in BarsChart for `from_bundle` to:
```
const transactionsBundleStats = useMemo(() => (transactions.txn_from_bundle.map((_, i) => getTxnBundleStats(i))), [getTxnBundleStats, transactions]);
```

src/features/SlotDetails/SlotTransactionsProvider.tsx
```diff
@@ -0,0 +1,189 @@
...
+        computeUnitStats.vote += consumedCus;
+      } else if (isBundle) {
+        computeUnitStats.bundle += consumedCus;
+      } else {
+        computeUnitStats.other += consumedCus;
+      }
+
+      // fee breakdown stats
+      feeBreakdownStats.fees += Number(getPaidTxnFees(transactions, i));
+      feeBreakdownStats.tips += Number(getPaidTxnTips(transactions, i));
```
amliu-jumptrading, line 150: Since we're already calculating paidTxnFees and paidTxnTips for the fee breakdown stats in the context, can we also calculate income once and return it as part of the context value?

We can have an income array that we accumulate values in
```
const income: number[] = [];
...
income[txnIdx] = paidFees + paidTips
```

We can have cheap access to the income values in each of the `IncomeDistributionCharts` and the income array used in `IncomeByPctTxns` is basically free since it got folded into the existing loop.

amliu-jumptrading (dismissed):

src/features/Overview/SlotPerformance/TransactionBarsCard/ChartTooltip.tsx
asuzuki-jumptrading: Nice! I'll remove the `txn_from_bundle` checks from a few other places too

src/features/SlotDetails/SlotTransactionsProvider.tsx
asuzuki-jumptrading: I'll add this but technically we'd have to store an extra array of length transactions.length for the income, whereas we previously calculated it as needed

asuzuki-jumptrading: Discussed offline -- we'll remove this so we don't store another array the size of the transactions length

amliu-jumptrading (approved):

PR #436, opened 2026-09-16
feat: smoothed now atom

asuzuki-jumptrading:
- Correct serverTimeNanos value to bigint (instead of number)
- Move the shreds adjusted now calculation into a global atom
- Use a ring buffer to keep track of rolling time diffs (between client now and server ts)


src/features/Overview/ShredsProgression/atoms.ts
```diff
@@ -192,7 +193,9 @@ export function createLiveShredsAtoms() {
 
         set(_liveShredsAtom, (prev) => {
           const slotRange = get(_slotRangeAtom);
-          const now = get(serverTimeMsAtom) ?? Date.now();
+          const delayedNow = getDelayedNow(
+            get(smoothedNowMsAtom) ?? Date.now(),
```
amliu-jumptrading, line 197: nit: can we use `performance.timeOrigin + performance.now()` instead of `Date.now()` for consistency with the `smoothedNowMsAtom`

src/api/worker/cache/shreds/shredsCalc.ts
```diff
@@ -102,9 +102,12 @@ export function createShredsCalc(getValidatorState: () => ValidatorState) {
    * Delete slots that completed before the chart x-axis starting time, or with dots outside visible x range
    * Update the min slot
    */
-  function deleteSlots(isStartup: boolean, serverTimeNanos?: number) {
+  function deleteSlots(isStartup: boolean, serverTimeNanos?: bigint) {
     const now =
-      serverTimeNanos == null ? Date.now() : serverTimeNanos / nsPerMs;
+      serverTimeNanos == null
+        ? Date.now()
```
amliu-jumptrading, line 108: nit: can we use `performance.timeOrigin + performance.now()` instead of `Date.now()` since the `smoothedNowMsAtom` uses that in it's calculation now

src/atoms.ts
```diff
@@ -859,10 +861,56 @@ export const [
...
+    requestAnimationFrame(function updateSmoothedNow(rafTime: number) {
+      if (rafTime - lastRafTime >= updateIntervalMs) {
+        store.set(refreshNowAtom);
+        lastRafTime = rafTime;
+      }
+      loop();
+    });
+  }
+
+  loop();
```
amliu-jumptrading, line 905: This loop runs on every page that renders a component importing `src/atoms.ts`, which is most of them, even if the shreds are not present on the page. It starts unconditionally at import and has no teardown mechanism, so there is wasted work even when the value is not used.

We could drop the loop and sample into the ring buffer each time a `server_time_nanos` message arrives from the websocket (updating an `avgTimeDiff` atom). Then we can calculate the `smoothedNowMs` as needed (calculated as `performance.timeOrigin + performance.now() - store.get(avgTimeDiff) / nsPerMs`). Each of the call sites currently already live within draw loops, so we get a fresh value each frame without managing any loop lifecycle.

A bonus if we do it this way is that we only need the `server_time_nanos` value to be a `bigint` when we subtract it to get the `timeDiff`. We can store the resulting diff in the ringbuffer as a `Number` and simplify our other map to not have to cast as `BigInt`.

asuzuki-jumptrading: Discussed offline -- we'll store the avgTime and weighted total in an atom that is updated only when `server_time_nanos` changes. We'll store the time diffs as numbers too

src/atoms.ts
```diff
@@ -859,10 +861,56 @@ export const [
...
+    const evicted = ringBuffer.push(timeDiff);
+
+    const total = prevTotal + timeDiff - (evicted ?? 0n);
+    prevTotal = total;
+    const avgTimeDiff = total / BigInt(ringBuffer.length);
+
+    const smoothedNowNs = nowNs - avgTimeDiff;
+    set(_smoothedNowNsAtom, smoothedNowNs);
+  });
+  const store = getDefaultStore();
```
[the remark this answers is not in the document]
asuzuki-jumptrading: will remove this with the other refactor

src/atoms.ts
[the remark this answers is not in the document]
asuzuki-jumptrading: Will remove this with the other refactor.
We want to be able to track offsets even if the components using this atom are not on the page.

src/atoms.ts
```diff
@@ -859,10 +861,56 @@ export const [
...
 })();
 
-export const serverTimeMsAtom = atom((get) => {
-  const serverTimeNanos = get(serverTimeNanosAtom);
-  if (serverTimeNanos == null) return undefined;
-  return Math.round(serverTimeNanos / nsPerMs);
+/**
+ * Get timestamp for "now", which is adjusted using an avg diff of server time and real now ts to smooth out server msg delays.
+ */
+export const smoothedNowNsAtom = (function () {
```
atsai-jumptrading: I thought from here we wanted to make it relative to validator start time so we didn't need to use bigints

atsai-jumptrading (approved):

